PR #9, opened 2017-02-27
Fix build by removing books from include paths in SUMMARY.adoc

sgilda:


jenmalloy:
@stianst Could this change break the GitBook build if it is merged?

stianst:
Yes, please don't merge this as it will break community builds. Why is this change being made?

sgilda:
@stianst : Please see my email with subject 'Issues with the keycloak-documentation repository' that I sent on February 27. Since you need the book folder in the path for the SUMMARY.adoc file, is it possible to update the Python script to generate the master.adoc file without the book folder in the include path?

PR #31, opened 2017-03-09
KEYCLOAK-4527 KEYCLOAK-4404

patriot1burke:
No import LDAP and Client Initiated Account Linking


stianst:
@jenmalloy FIY this is for Keycloak 3.x only

jenmalloy:
@stianst Thanks!

PR #38, opened 2017-03-14
Fixing sentence

pedroigor:
@jenmalloy, you suggested this change, right ?


jenmalloy:
@pedroigor, yes, thank you!

PR #55, opened 2017-03-17
Docker Protocol Docs

josh-cain:
Combines server_admin_guide #59 and securing_apps_guide #55

Also included a section for the quickstart .zip to help clarify some of the confusion on KEYCLOAK-3592


securing_apps/topics/docker/quickstart-zip.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+WARNING: This installation method is meant to be an easy way to get a docker registry authenticating against a keycloak server.  It is intended for development purposes only and should never be used in a production or production-like environment.
+
+The zip file installation mechanism provides a quickstart for developers who want to understand how the keycloak server can interact with the docker registry.  In order to configure the quickstart:
+
+ 1. Begin by creating a client configuration.  At this point you won't have a docker registry - the quickstart will take care of that part.
+ 1. Choose the "Quickstart" option from the installation tab and download the .zip file
+ 1. Unzip the archive to the desired location, and open the directory.
+ 1. Start the docker registry with `docker-compose up`
+ 1. Configure the realm in which the docker client is configured to use HTTP/Basic authentication as the browser auth flow
```
mstruk: This part about HTTP Basic Authentication should be more explicit - i.e.:

1. In Admin Console realm configuration under *Authentication* / *Flows* tab click *New* to create a new Authentication flow. Give it some alias e.g. 'docker auth'. For *Top Level Flow Type* select 'generic'.
1. Click on *Add execution* and select a provider called 'HTTP Basic Authentication'.
1. Make sure the execution is set to *Required*.
1. Go to *Bindings* tab, and change 'Browser Flow' to just created 'docker auth'.

securing_apps/topics/docker/quickstart-zip.adoc
```diff
@@ -0,0 +1,22 @@
+[[_docker_quickstart_zip]]
+
+=== Docker Registry Quickstart Zip
+
+WARNING: This installation method is meant to be an easy way to get a docker registry authenticating against a keycloak server.  It is intended for development purposes only and should never be used in a production or production-like environment.
+
+The zip file installation mechanism provides a quickstart for developers who want to understand how the keycloak server can interact with the docker registry.  In order to configure the quickstart:
+
+ 1. Begin by creating a client configuration.  At this point you won't have a docker registry - the quickstart will take care of that part.
```
mstruk: Let's instruct the user to first create a new realm to avoid issues after setting Browser Flow to custom one that only contains 'HTTP Basic Authentication'.

securing_apps/topics/docker/quickstart-zip.adoc
mstruk: Note:  Authenticator to use is now 'Docker Authenticator', and not 'HTTP Basic Authentication'.

josh-cain:
@mstruk Updated.  Also changed the 'quickstart' verbiage out to correspond to accompanying PR.

stianst:
@cainj13 I'll try to find time to dig into this PR in more detail, but it seems that for some reason you are using the browser based authentication flow and some custom authenticators? That makes little sense to me as the Docker client is not browser based. Why is it not just using the direct grant flow?

thomasdarimont:
This looks awesome! I hope the will there be a blog post about authenticating 
against Keycloak when using a docker registry :)

josh-cain:
@stianst Yeah, you're right.  This kind of resource owner credentials -> token exchange would probably fit better under that type of flow.  Would switching to that flow suffice?  There is discussion on KEYCLOAK-3592 of a new flow entirely...

josh-cain:
@thomasdarimont There most certainly will be ;)

server_installation/topics/profiles.adoc
```diff
@@ -34,6 +34,7 @@ The features that can be enabled and disabled are:
 * Authorization - authorization services
 * Impersonation - ability for admins to impersonate users
 * Script - write custom authenticators using JavaScript
+* Docker - use docker-v2 authentication protocol
```
vmuzikar: Should we mention that the docker profile is disabled by default in the community (unlike the all other profiles which are enabled)?

mstruk:
We should also add instructions on how to activate docker-v2 feature.

josh-cain:
@mstruk Added instructions and squashed to single commit

stianst (changes requested):
Fixed some issues in summary and I made two changes here https://github.com/cainj13/keycloak-documentation/pull/1:
* Merge to a  single page. I think this is better than the 3 different pages as there's not much on each page.
* Link to profile page for details on how to enable and also updated the profiles page. I will improve the profiles page later and also want to add an option to configure the profile in standalone.xml instead of a different properties file (https://issues.jboss.org/browse/KEYCLOAK-4994).

One addition that would be nice is a brief intro to the 3 different options in the first section.

PR #56, opened 2017-03-20
KEYCLOAK-4335: documentation for x509 client certificate authentication

brat000012001:
Documentation for x509 client cert authentication (as per Stian's request).

JIRA: KEYCLOAK-4610


stianst:
@jenmalloy FIY this is not for 7.1

PR #93, opened 2017-04-03
Fixed typos about technical term

tnorimat:
Fixed some typos about technical term in Compromised Access Codes  and Session and Token Timeouts.


stianst:
Thanks

PR #137, opened 2017-05-20
Add asciidoc-admonition-icons GitBook plugin.

msavy:
I recently wrote a plugin that reintroduces the missing font icons for AsciiDoc admonition blocks. I made some enhancements a few days ago to support GitBook 2.x.x.

Please have a play around and see what you think :-). A bunch of customisations are available if you're so inclined: https://github.com/msavy/gitbook-plugin-asciidoc-admonition-icons#configuration

LMK if I should apply it to any other branches.

Here are some screenshots:

image

image

image

### DIFF
--- a/book.json
+++ b/book.json
@@ -7,7 +7,8 @@
     "toggle-chapters",
     "ungrey",
     "splitter",
-    "ga"
+    "ga",
+    "asciidoc-admonition-icons"
   ],
   "pluginsConfig": {
     "ga": {
--- a/server_development/book.json
+++ b/server_development/book.json
@@ -6,7 +6,8 @@
   "plugins": [
     "toggle-chapters",
     "ungrey",
-    "splitter"
+    "splitter",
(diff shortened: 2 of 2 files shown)


stianst:
Thanks @msavy we're actually thinking about ditching Gitbook though and go full asciidoctor. If we don't do that and stick with Gitbook I'll merge this.

msavy:
I just moved apiman over to GitBook; took me an age. Maybe I should have
gone with vanilla asciidoctor 😨..!

stianst:
The problem we're facing is that we have to run a Python script to convert "gitbook friendly ascii" to RH tools friendly product docs. That's the main reason we're doing the change.

My plan is asciidoctor via a Maven plugin and then use Travis CI to publish to Github pages automatically.

stianst:
Closing this as we've decided to go with asciidoctor instead of Gitbook

PR #147, opened 2017-06-06
KEYCLOAK-4980 SAML adapter should return 401 when unauthenticated Aja…

hokuda:
…x client accesses
to comply with the request in the PR [4192](https://github.com/keycloak/keycloak/pull/4192)


hmlnarik:
Looks good to me

PR #149, opened 2017-06-15
Usage for Spring metadata annotation @KeycloakConfiguration added

hendrikebbers:
Spring security sample use @KeycloakConfiguration now (see https://github.com/keycloak/keycloak/pull/4199) and a tip for custom / more complex scenarios is added.


sebastienblanc:
:+1: 
@stianst Can you merge ?

PR #191, opened 2017-09-05
Convert frames.html into non-frame global navigation bar

ASzc:
This is done using the "docinfo.html" mechanism of asciidoctor to include css in the header, and html in the body/footer of each page.

Symlinks are used to direct each build to use the common docinfo files located in `aggregation/`.

Since I was forced to use fixed positioning, the CSS isn't perfect (drops a navigation element in some scenarios), but I've tried to make it work with different screen sizes.


stianst:
Excellent. Thanks for sorting this out :)

PR #201, opened 2017-09-14
Matthew Helmke edits for Red Hat writing assessment

matthewhelmke:
I have not read any style guide in use for files in the repo, so I made a decision that could be controversial along with the text edits. Specifically, I chose to narrow the margins within the file for easier viewing while editing across all text editors. The right margin does not affect the Markdown processors  I have used, all of which ignore CRs and format text into paragraphs based on an empty white line. I hope that is also the case here with asciidoc.

Other changes include:
* Removing double spaces between sentences
* Changing passive voice to active
* Expanding the reasoning given for limiting use of the master realm
* Changing some phrasing in favor of expressions that I thought sounded closer to a standard in technical documentation, such as replacing "king of all realms" to "the highest level in the hierarchy of realms"

### DIFF
--- a/server_admin/topics/realms/master.adoc
+++ b/server_admin/topics/realms/master.adoc
@@ -1,15 +1,23 @@
 
 === The Master Realm
 
-When you boot {project_name} for the first time a pre-defined realm is created for you.  This initial realm is called
-the _master_ realm and is the king of all realms.  Admins in this realm have permissions to view and manage any
-other realm created on the server instance.  When you define your initial admin account, you are creating an account in the _master_ realm.
-Your initial login to the admin console will also be through the _master_ realm.
+When you boot {project_name} for the first time {project_name} creates a
+pre-defined realm for you. This initial realm is the _master_ realm. It is the
+highest level in the hierarchy of realms. Admin accounts in this realm have
+permissions to view and manage any other realm created on the server instance.
+When you define your initial admin account, you create an account in the
+_master_ realm. Your initial login to the admin console will also be via the
+_master_ realm.
 
-It is recommended that you do not use the _master_ realm to manage the users and applications in your organization.  Keep the _master_ realm
-as a place for _super_ admins to create and manage the realms in your system.  This keeps things clean and organized.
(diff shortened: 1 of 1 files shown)


patriot1burke:
If you want to do minor things like reducing line columns, then please do that sort of stuff in a separate PR.  Its often hard to distinguish actual edits from formatting.

BTW, why are you removing double spaces between sentences?  That's something I learned back in type class in high school and is pretty much a standard in any book you read.

matthewhelmke:
Your comment on the reduction of columns in a separate PR is quite reasonable. I'll keep that in mind for the future.

One space following a period at the end of a sentence is the standard and has been for years, although those of use who learned to type using actual typewriters (including myself) were taught to use two spaces. So, the thought is a reasonable one. In the digital era, spacing is dealt with by typesetting software, not manually. The extra space is not needed and can cause problems in some circumstances. [https://www.google.com/search?q=how+many+spaces+after+a+period](https://www.google.com/search?q=how+many+spaces+after+a+period).

From the IBM Style Guide, page 59: "Use a period at the end of a sentence. After a period, use the spacing that your authoring tool provides. If you insert spacing manually, insert only one space after a period." Other common/major style guides and manuals concur, such as:
- [The Microsoft Manual of Style](https://www.safaribooksonline.com/library/view/microsoft-manual-of/9780735669833/ch09s01.html)
- [The Chicago Manual of Style](http://www.chicagomanualofstyle.org/qanda/data/faq/topics/OneSpaceorTwo.html)
- [The Modern Language Association MLA Handbook](https://style.mla.org/2016/07/27/number-of-spaces-after-period/)

stianst:
@matthewhelmke an early welcome to team Keycloak :)

matthewhelmke:
Thank you, @stianst!

stianst:
Changes looks good to me, except "Red Hat recommends" part. I don't think we should state that in the community/Keycloak version of the docs.

matthewhelmke:
Sounds good. I'll update momentarily.

stianst:
Thanks

PR #203, opened 2017-09-18
fix ldap documentation

muehlburger:


stianst:
Thanks. Always nice to get a PR instead of a bug report :)

PR #211, opened 2017-10-03
Fix spelling for HTTP requests

muehlburger:
This fix corrects spelling issues for HTTP requests.


stianst:
It's "a HTTP request", not "an HTTP request". The other edits are sound though.

muehlburger:
@stianst I wasn't sure either and looked it up in the Spec ("[Hypertext Transfer Protocol -- HTTP/1.1](https://tools.ietf.org/html/rfc2616#section-1.3)"). They use "an", literally "An HTTP request message ...".

stianst:
You might be right. When considering it some more I do for sure say "An HTTP..". I've asked our docs team for some input, so should know shortly.

matthewhelmke:
The general rule is that abbreviations that start with a letter that has an initial vowel sound are preceded by "an." In this case, "H" is said "aytch," and should therefore use "an."

stianst:
@matthewhelmke thanks

muehlburger:
@matthewhelmke @stianst thanks

PR #219, opened 2017-10-23
Fix: Incorrect number of closing parenthesis

ghost:


matthewhelmke:
Thank you! We appreciate you noticing and fixing this.

PR #220, opened 2017-10-23
Fix: s/is can/can/

ghost:


matthewhelmke:
Thank you! We appreciate you catching and fixing this.

PR #221, opened 2017-10-24
add the missing binding step to enable the x509 client authentication

wei-lee:
steps

When I try to enable X509 client authentication by following the documents, it didn't work initially. It is because the document is missing one important step to bind the X509 authentication flow to the corresponding flow.

This PR will update the X509 authentication doc to add the missing binding step.

Please let me know if there is anything else I need to do.

Thanks.


matthewhelmke:
Thank you. We are in the middle of planning meetings and this has delayed our ability to review. We will review this soon, probably next week.

wei-lee:
@matthewhelmke no problem at all.

stianst:
@mposolda can you review this please?

mposolda (approved):

PR #223, opened 2017-11-01
Fix: s/regisration/registration/

czosel:


matthewhelmke:
Thank you! We appreciate you catching and fixing this.

PR #224, opened 2017-11-01
Adding a statement about the need to add the 'username' of the newly created user in the host-slave.xml

priyanka-nag:
The 'Setup Slave Connection to Domain Controller' topic of the Server Installation and Configuration guide doesn't mention about the need to add the username of the newly created user to the xml file of the slave. Without this, the connection cannot be complete. I have added the step for the same.

### DIFF
--- a/server_installation/topics/operating-mode/domain.adoc
+++ b/server_installation/topics/operating-mode/domain.adoc
@@ -269,6 +269,13 @@ Now cut and paste the secret value into the _.../domain/configuration/host-slave
                  </server-identities>
 ----
 
+You will also need to add the _username_ of the created user in the _.../domain/configuration/host-slave.xml_ file:
+
+[source,xml]
+----
+     <remote security-realm="ManagementRealm" username="admin">
+----
+
 ===== Run the Boot Scripts
 
 Since we're simulating a two node cluster on one development machine, you'll run the boot script twice:


matthewhelmke:
Thank you. We have some planning meetings this week and this has delayed our review process. We definitely appreciate you catching this and submitting a fix!

stianst:
@mstruk can you review this one please?

mstruk:
@stianst Ok, I'll review.

mstruk (changes requested):
There is no explanation given - a reason - for this additional step.

mstruk:
I also have a bigger issue with the whole section. When I try this domain mode example I come to a wall when I go to localhost:8080/auth and there is no form for creating a user. Trying `bin/add-user-keycloak.sh --domain -u admin -r master -p admin` creates /domain/configuration/keycloak-add-user.json which never gets processed when server is started. I've no idea how to run a workable example here. Some additional steps are apparently missing.

priyanka-nag:
@Paramvir-JIndal, can you check on @mstruk's comment and help me with this one?

Paramvir-JIndal:
@mstruk,
To answer your first question, we need to add the **username** in the xml file cause this is required to build the connection between the mater and the slave. This is the same username created using _add-user.sh_ script. The secret value that we add into the `…​/domain/configuration/host-slave.xml` file needs the associated username as well along with it for the connection to be established. 

Coming to the second point now. I am looking at - https://access.redhat.com/documentation/en-us/red_hat_single_sign-on/7.1/html/server_administration_guide/server_initialization. If you see the documentation, you will need to create the **configuration** directory under the `$RH-SSO_HOME/domain/servers/server-one` path. Also, the command you need to use here is `$RH-SSO_HOME/bin/add-user-keycloak.sh --sc $RH-SSO_HOME/domain/servers/server-one/configuration -r master -u <username> -p <password>`. Now when the _keycloak-add-user.json_ is created, it will be processed fine when the server is started.

stianst (approved):

PR #236, opened 2017-11-14
Change curl PAT example to use form parameters instead of BASIC

pedroigor:


matthewhelmke:
Thanks, Pedro!

matthewhelmke:
@stianst I know this was discussed on the mailing list. Are we ready to merge?

PR #237, opened 2017-11-15
created contributors guide and edited existing related files

matthewhelmke:
The templates and other helps still need some clean up. This seems like a good MVD moment to push, while keeping that in mind.


stianst:
Looks good to me. Would be good to have a link to some abbreviated version of the IBM style guide if that exists.

stianst:
If that doesn't exist maybe we could create some short version of what you think is important?

matthewhelmke:
It is available in Safari (www.safaribooksonline.com), but that is the only online access I have ever found.

We have a `styleguide.adoc` file in the `internal_resources` directory. I'll take a look at it and see how it could be used to communicate a subset of the most important points. I'll prepare that for a later PR.

PR #243, opened 2017-11-24
Add note about the behavior of X.509 Authentication for Key Usage

abstractj:
The behavior of *Key Usage* and *Extend Key Usage* is not totally clear
at the docs.

This change clarifies a little bit about when these extensions will be
ignored or mandatory.


matthewhelmke:
Thank you. We appreciate your help.

PR #250, opened 2017-11-29
KEYCLOAK-5916 : add warning for Spring Boot deployed as WARs

sebastienblanc:


stianst:
@sebastienblanc isn't the web.xml optional in servlet 3.0?

sebastienblanc:
@stianst yes but how should one specify `security-constraints`  then ?

PR #252, opened 2017-11-30
KEYCLOAK-5950 - fixed cross references

matthewhelmke:
Cross references need to use a different markup syntax than links. I did not realize that before now. This fixes the broken cross reference links in the doc.


matthewhelmke:
@hmlnarik This has been tested locally and works. Please review.

hmlnarik (approved):
Thanks for the fix!

PR #254, opened 2017-12-01
Admincli edits

leswilliams44:
Completed the first pass on Chapter 19, Admin-CLI. Main changes include:
* changing tense to present
* writing in the second person, imperative mood
* adding non-visible AsciiDoc syntax
* following IBM Style Guide conventions
* adding articles and nouns where necessary for clarification

### DIFF
--- a/server_admin/topics/admin-cli.adoc
+++ b/server_admin/topics/admin-cli.adoc
@@ -106,7 +106,7 @@ This approach maintains an authenticated session between the [command]`kcadm` co
 
 The second approach only authenticates each command invocation for the duration of that invocation. This approach increases the load on the server and the time spent with roundtrips obtaining tokens. The benefit of this approach is not needing to save any tokens between invocations, which means nothing is saved to disk. This mode is used when the [command]`--no-config` argument is specified.
 
-For example, when performing an operation, we specify all the information required for authentication.
+For example, when performing an command, we specify all the information required for authentication.
 [options="nowrap"]
 ----
 $ kcadm.sh get realms --no-config --server http://localhost:8080/auth --realm master --user admin --password admin
@@ -136,9 +136,9 @@ Consider avoiding storing secrets inside a configuration file, even though it is
 
 === Basic operations and resource URIs
 
-The Admin CLI allows you to generically perform CRUD operations against Admin REST API endpoints in with additional commands that simplify performing certain tasks.
+The Admin CLI allows you to generically perform CRUD operations against Admin REST API endpoints with additional commands that simplify performing certain tasks.
 
-The main usage pattern is listed below, where the operations [command]`create`, [command]`get`, [command]`update`, and [command]`delete` are mapped to the HTTP verbs POST, GET, PUT, and DELETE, respectively.
+The main usage pattern is listed below, where the [command]`create`, [command]`get`, [command]`update`, and [command]`delete` commands are mapped to the HTTP verbs `POST`, `GET`, `PUT`, and `DELETE`, respectively.
 [options="nowrap"]
 ----
 $ kcadm.sh create ENDPOINT [ARGUMENTS]
@@ -165,7 +165,7 @@ SERVER_URI/admin/realms
 
 There is also a [filename]`serverinfo` endpoint, which is treated the same way because it is independent of realms.
 
-When you authenticate as a user with realm-admin powers, you might need to perform operations on multiple realms. In that case, specify the [command]`-r` option to tell explicitly which realm the operation should be executed against. Instead of using REALM as specified via the [command]`--realm` option of [command]`kcadm.sh config credentials`, the TARGET_REALM is used.
+When you authenticate as a user with realm-admin powers, you might need to perform commands on multiple realms. In that case, specify the [command]`-r` option to tell explicitly which realm the command should be executed against. Instead of using [filename]`REALM` as specified via the [command]`--realm` option of [command]`kcadm.sh config credentials`, the [filename]`TARGET_REALM` is used.
 
 [options="nowrap"]
 ----
@@ -205,14 +205,14 @@ $ kcadm.sh update realms/demorealm -s enabled=false
 
 That method only updates the [command]`enabled` attribute to `false`.
 
-By default, the [commamd]`update` operation first performs a [command]`get` and then merges the new attribute values with existing values. This is the preferred behavior. In some cases, the endpoint may support [command]`PUT` but not [command]`GET`. You can use [command]`-n` option to perform a "no-merge" update, which performs a [command]`PUT` without first doing a [command]`GET`.
+By default, the [commamd]`update` command first performs a [command]`get` and then merges the new attribute values with existing values. This is the preferred behavior. In some cases, the endpoint may support the [command]`PUT` command but not the [command]`GET` command. You can use the [command]`-n` option to perform a "no-merge" update, which performs a [command]`PUT` command without first running a [command]`GET` command.
 
 
 === Realm operations
 
 ==== Creating a new realm
 
-Use the [command]`create` operation on the `realms` endpoint to create a new enabled realm, and set the attributes to `realm` and `enabled`.
+Use the [command]`create` command on the `realms` endpoint to create a new enabled realm, and set the attributes to `realm` and `enabled`.
 [options="nowrap"]
 ----
 $ kcadm.sh create realms -s realm=demorealm -s enabled=true
@@ -323,7 +323,7 @@ $ kcadm.sh update realms/demorealm -s registrationAllowed=true -s registrationEm
 
 ==== Listing the realm keys
 
-Use the [command] `get` operation on the [filename]`keys` endpoint of the target realm.
+Use the [command]`get` operation on the [filename]`keys` endpoint of the target realm.
 [options="nowrap"]
 ----
 $ kcadm.sh get keys -r demorealm
@@ -332,7 +332,7 @@ $ kcadm.sh get keys -r demorealm
 
 ==== Generating new realm keys
 
-. Get the `ID` of the target realm before adding a new RSA-generated key pair.
+. Get the ID of the target realm before adding a new RSA-generated key pair.
 +
 For example:
 +
@@ -356,7 +356,7 @@ $ kcadm.sh create components -r demorealm -s name=rsa-generated -s providerId=rs
 ----
 c:\> kcadm create components -r demorealm -s name=rsa-generated -s providerId=rsa-generated -s providerType=org.keycloak.keys.KeyProvider -s parentId=959844c1-d149-41d7-8359-6aa527fca0b0 -s "config.priority=[\"101\"]" -s "config.enabled=[\"true\"]" -s "config.active=[\"true\"]" -s "config.keySize=[\"2048\"]"
 ----
-. Set the `parentId` attribute to the value of target realm's `ID`.
+. Set the `parentId` attribute to the value of target realm's ID.
 +
 The newly added key should now become the active key as revealed by [command]`kcadm.sh get keys -r demorealm`.
 
@@ -380,7 +380,7 @@ $ kcadm.sh create components -r demorealm -s name=java-keystore -s providerId=ja
 c:\> kcadm create components -r demorealm -s name=java-keystore -s providerId=java-keystore -s providerType=org.keycloak.keys.KeyProvider -s parentId=959844c1-d149-41d7-8359-6aa527fca0b0 -s "config.priority=[\"101\"]" -s "config.enabled=[\"true\"]" -s "config.active=[\"true\"]" -s "config.keystore=[\"/opt/keycloak/keystore.jks\"]" -s "config.keystorePassword=[\"secret\"]" -s "config.keyPassword=[\"secret\"]" -s "config.alias=[\"localhost\"]"
 ----
 . Make sure to change the attribute values for `keystore`, `keystorePassword`, `keyPassword`, and `alias` to match your specific keystore.
-. Set the `parentId` attribute to the value of target realm's `ID`.
+. Set the `parentId` attribute to the value of target realm's ID.
 
 
 ==== Making the key passive or disabling the key
@@ -391,7 +391,7 @@ c:\> kcadm create components -r demorealm -s name=java-keystore -s providerId=ja
 ----
 $ kcadm.sh get keys -r demorealm
 ----
-. Use the key's `providerId` attribute to construct an endpoint URI: [filename]`components/PROVIDER_ID`.
+. Use the key's `providerId` attribute to construct an endpoint URI, such as [filename]`components/PROVIDER_ID`.
 . Perform an [command]`update`.
 +
 For example, on:
@@ -494,7 +494,7 @@ $ kcadm delete events
 
 ==== Flushing the caches
 
-. Use the [command]`create` operation and one of the following endpoints: [filename]`clear-realm-cache`, [filename]`clear-user-cache`, or [filename]`clear-keys-cache`.
+. Use the [command]`create` command and one of the following endpoints: [filename]`clear-realm-cache`, [filename]`clear-user-cache`, or [filename]`clear-keys-cache`.
 . Set `realm` to the same value as the target realm.
 +
 For example:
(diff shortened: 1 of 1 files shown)


matthewhelmke:
Thank you, @leswilliams44 . I appreciate your attention to detail and kind assistance. Looks good to me.

matthewhelmke:
@stianst This is ready to merge. I just wanted to make sure you had a chance to have a dev team member review, if desired. The changes were all linguistic and should not impact techinical content.

stianst:
@mstruk do you want to check these edits?

mstruk:
@stianst Yes, I'll check today.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
+====
+To avoid repetition, the rest of this document only gives Windows examples in places where the difference in the CLI is more than just in the [command]`kcadm` command name.
+====
+
+
+=== Using the Admin CLI
+
+The Admin CLI works by making HTTP requests to Admin REST endpoints. Access to them is protected and requires authentication.
+
+//Should there be a link for the REST API documentation?
```
mstruk: It would be nice to have a link. The reason I have not included one is most likely that I wasn't sure we have a canonical url that won't change in the future. By avoiding the link we avoid a possibility of having a broken link.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
+[options="nowrap"]
+----
+c:\> kcadm config credentials --server http://localhost:8080/auth --realm demo --user admin --client admin
+c:\> kcadm create realms -s realm=demorealm -s enabled=true -o
+c:\> kcadm create clients -r demorealm -s clientId=my_client -s "redirectUris=[\"http://localhost:8980/myapp/*\"]" -i > clientid.txt
+c:\> set /p CID=<clientid.txt
+c:\> kcadm get clients/%CID%/installation/providers/keycloak-oidc-keycloak-json
+----
+
+. In a production environment, you access {project_name} with `https:` to avoid exposing tokens to network sniffers. If a server's certificate is not issued by one of the trusted certificate authorities (CAs) that are included in Java's default certificate truststore, prepare a [filename]`truststore.jks` file and instruct the Admin CLI to use it.
```
mstruk: I would change 'you access' to 'you have to access'.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
+----
 
 
 === Authenticating
 
-When logging in with `Admin CLI` you specify a server endpoint url, and a realm. Then you specify a username,
-or alternatively you can specify only a client id, which will result in special, so called 'service account' being used. In the first case,
-a password for the specified user has to be used at login. In the latter case there is no user password - only client secret.
-Alternatively, `Signed JWT` can be used.
+When you log in with the Admin CLI, you specify a server endpoint URL and a realm, and then you specify a username. Another option is to specify only a client ID, which results in using a special "service account". When you log in using a username, you must use a password for the specified user. When you log in using a client ID, you only need the client secret, not the user password. You could also use [command]`Signed JWT` instead of the client secret.
```
mstruk: Same here as with Client Registration CLI, it would be better to use 'client id' instead of 'client ID' as it's a colloquial reference to 'clientId' attribute, rather than 'id' attribute.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
 
-    $ kcadm.sh config credentials --server http://localhost:8080/auth --realm master --user admin --password admin
+The second approach only authenticates each command invocation for the duration of that invocation. This approach increases the load on the server and the time spent with roundtrips obtaining tokens. The benefit of this approach is not needing to save any tokens between invocations, which means nothing is saved to disk. This mode is used when the [command]`--no-config` argument is specified.
 
-This approach maintains an authenticated session between `kcadm` command invocations by saving the obtained access token, and
-associated refresh token, possibly other secrets as well in a private configuration file. See <<_working_with_alternative_configurations, next chapter>> for more info on configuration file.
+For example, when performing an command, we specify all the information required for authentication.
```
mstruk: type: 'an command' -> 'a command'

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
 
+Run the [command]`kcadm.sh config credentials --help` command for more information about starting an authenticated session.
 
 
 [[_working_with_alternative_configurations]]
 === Working with alternative configurations
 
-By default, `Admin CLI` automatically maintains a configuration file called `kcadm.config` located under user's home
-directory - it's full pathname is `$HOME/.keycloak/kcadm.config` (on Windows it's `%HOMEPATH%\.keycloak\kcadm.config`).
+By default, the Admin CLI automatically maintains a configuration file called [filename]`kcadm.config` located under the user's home directory, which is the full path name [filename]`$HOME/.keycloak/kcadm.config`. On Windows, the full path name is [filename]`%HOMEPATH%\.keycloak\kcadm.config`.
```
mstruk: I think instead of 'which is the full path name ...' it should be 'with the full path name ...'

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
-NOTE: It's best to perform operations tied to a single config file from a single thread.
+[NOTE]
+====
+It is best to perform operations tied to a single configuration file from a single thread.
+====
 
-Make sure to not make config file visible to other users on the system as it contains access tokens, and secrets that should be kept private.
-By default the ~/.keycloak directory and its content will be automatically created with proper access limits. However if directory will
-exist already with non-default permissions, those will not be updated.
+Make sure you do not make the configuration file visible to other users on the system. It contains access tokens and secrets that should be kept private. By default, the [filename]`~/.keycloak` directory and its content are created automatically with proper access limits. If the directory already exists with non-default permissions, those access tokens and secrets will not be updated.
```
mstruk: The last sentence is semantically incorrect. If the directory already exists then permissions won't be updated (rather than tokens and secrets).
I suggest: 'If the directory already exists, its permissions won't be modified.'

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
 
-You may want to avoid storing any secrets at all inside a config file for the price of less convenience and having to do more token requests.
-In that case you can use `--no-config` option with all your commands. In that case you will have to specify all the authentication info needed by
-`config credentials` command with each `kcadm` invocation.
+Consider avoiding storing secrets inside a configuration file, even though it is less convenient and you have to make more token requests. If you choose to not store secrets, you can use the [command]`--no-config` option with all your commands and specify all the authentication information needed by the [command]`config credentials` command with each [command]`kcadm` invocation.
```
mstruk: Using 'Consider' here sounds as a suggestion - a hint that it's a better option. I don't think it is a better option. This support is only here for the extra paranoid, and it will probably end up being used by only a very tiny percentage of users. In fact another use-case is when you only need to perform a single operation, and it's more efficient to do it in one single step.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
+[options="nowrap"]
+----
+$ kcadm.sh create ENDPOINT [ARGUMENTS]
+$ kcadm.sh get ENDPOINT [ARGUMENTS]
+$ kcadm.sh update ENDPOINT [ARGUMENTS]
+$ kcadm.sh delete ENDPOINT [ARGUMENTS]
+----
 
-Main usage pattern is the following:
+ENDPOINT is a target resource URI and can either be absolute (starting with `http:` or `https:`) or relative, used to compose an absolute URL of the following format.
```
mstruk: I think this sentence should end with a colon - ':'

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
-Creating a new realm::
-
-To create a new enabled realm use `create` operation on `realms` endpoint, and set attributes `realm` and `enabled`:
-
-    $ kcadm.sh create realms -s realm=demorealm -s enabled=true
-
-Realm is not enabled by default. By enabling it, it can be used for authentication immediately.
-
-A description for a new object can be in JSON format as well:
+==== Creating a new realm
```
mstruk: My concern around replacing 'Subtitle::' notation with '==== Subtitle' notation is that the second one results in all these subtitles being displayed in navigation in the left panel. The number of these titles is significant, and they will dominate the whole document's navigation panel, making it also aesthetically displeasing. I was unable to find a way to suppress the fourth level of depth in navigation, and that is the reason 'Subtitle::' notation was used (using labels instead of subtitles).

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
+[options="nowrap"]
+----
+$ kcadm.sh get realms --fields realm --format csv --noquotes
+----
 
-You can even display the result as comma separated values:
+==== Getting a specific realm
 
-    $ kcadm.sh get realms --fields realm --format csv --noquotes
+You append an ID to a collection URI to get an individual item of a collection, as is common for REST web services.
```
mstruk: Actually this sentence is dubious, and was dubious in the first place. It's an aberration that for realm we don't use 'id' field, but 'realm' field. Saying we add a realm id is thus misleading. 
I suggest: 'You append a realm name to a collection URI to get an individual realm.'

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
-Then, create a new role by using client's `id` attribute to construct an endpoint uri - `clients/ID/roles`.
+==== Creating a client role
 
+. Identify the client first and then use the [command]`get` command to list available clients when creating a client role.
++
+[options="nowrap"]
+----
+$ kcadm.sh get clients -r demorealm --fields id,clientId
+----
+. Create a new role by using the client's ID attribute to construct an endpoint URI, such as [filename]`clients/ID/roles`.
```
mstruk: Throughout this document I see consistent replacement of 'id' with ID, but at this point here it really makes it painfully obvious to me that it's just wrong. JSON is case sensitive, and has no notion of primary key. Saying 'client's ID attribute' is just wrong. It should really be 'client's 'id' attribute'.

What is the rationale for this consistent replacement of 'id' with ID?
I see in the rest of the document, that where formulation is "'id' attribute" it was not changed into ID. In fact it looks like all occurences of ID can be replaced with "'id' attribute". That would make everything more clear and consistent.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
 
-    $ kcadm.sh get-roles -r demorealm --cclientid realm-management
 
+==== Getting a specific realm role
 
-
-Getting a specific realm role::
-
-Use `get` command, and role `name` to construct an endpoint uri for a specific realm role - `roles/ROLE_NAME`
+Use the [command]`get` command and the role [filename]`name` to construct an endpoint UIR for a specific realm role: [filename]`roles/ROLE_NAME`, where [filename]`user` is the name of the existing role.
```
mstruk: Typo: 'endpoint UIR' -> 'endpoint URI'

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
 
+. Determine the ID of the composite client role by using the [command]`get-roles` command.
++
+For example:
++
+[options="nowrap"]
+----
+$ kcadm.sh get-roles -r demorealm --cclientid test-client --rolename operations
+----
+. Assume that there is a client with client ID of [filename]`test-client`, a client role called [filename]`support`, and another client role, which becomes a composite role, that has an ID of `"fc400897-ef6a-4e8c-872b-1581b7fa8a71", "name":"operations"`.
```
mstruk: There's something wrong with the ending of this sentence. 

I suggest:
'..., and another client role called [filename]`operations`, which becomes a composite role, that has an ID of  "fc400897-ef6a-4e8c-872b-1581b7fa8a71"'.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
++
+In this example, [filename]`operations` is the target composite role, and the role's ID precedes [filename]`operations`.
```
mstruk: I don't understand the 'precedes' part of this sentence nor what example this sentence refers to. Also seems like it doesn't fit right with the next sentence.

I think this sentence can safely be removed, as all the relevant information is given in the previous sentence.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
+$ kcadm.sh remove-roles --rname testrole --rolename user -r demorealm
+----
 
-    $ kcadm.sh get-roles -r demorealm --cclientid test-client --rolename operations
 
-Let's assume that there exists a client with "clientId": 'test-client', a client role called 'support', and another client role - that will become composite role - that has an "id": "fc400897-ef6a-4e8c-872b-1581b7fa8a71", "name":"operations".
+==== Adding client roles to a realm role
 
-In this example 'operations' is our target composite role, and we just got its `id`.
+You can create or modify a composite realm role.
```
mstruk: I think reworked text makes this sentence redundant. Originally it is simply an intro into a recipe how to do that. When it stands alone like this it seems to suggest that you can create or modify, but maybe not delete and read, which is not true. I suggest removing it.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
+----
+$ kcadm.sh add-roles -r demorealm --rname testrole --cclientid realm-management --rolename create-client --rolename view-users
+----
 
 
-Afterwards all the roles of a composite role can be listed by using `get-roles --all`. For example:
+==== Adding client roles to a client role
 
-    $ kcadm.sh get-roles --rid fc400897-ef6a-4e8c-872b-1581b7fa8a71 --all
+You can create or modify a composite client role.
```
mstruk: Again, this sentence is redundant.

server_admin/topics/admin-cli.adoc
```diff
@@ -1,1320 +1,1702 @@
...
+----
+$ kcadm.sh create users -r demorealm -s username=testuser -s enabled=true
+----
 
 
-Updating a user::
+==== Listing users
 
-Use `update` operation with the same endpoint uri as for getting a specific user. For example on Linux:
+Use the [filename]`users
```
mstruk: Second sentence is incorrect. It is not admin who will change the password and who is the intended audience of this document. Rather it is one of the users who will have to change their password when they log in the next time.

I suggest: 'The target user will have to change the password the next time they log in.'

mstruk: This sentence doesn't sound quite right. Maybe a slight rewording:

'You can use the [command]`--userid` to specify the user by the value of [command]`id` attribute.'

mstruk: Last sentence should rather be: "The target user will have to change the password the next time they log in."

mstruk: I think it sounds better to say: "... on the [command]`realm management` client ...".

mstruk: I think it sounds better to say: "... on the [command]realm management client ...".

mstruk: Typo - repeated word: 'If you want want'

mstruk (changes requested):

mstruk:
For some reason line numbers on my suggestions are invalid. That sucks big time.

server_admin/topics/admin-cli.adoc
matthewhelmke: Sounds reasonable. Is it standard practice in this project to leave the comment in the file and replace it in the future when/if we have a canonical URL for the REST API documentation, or does it need to be removed?

server_admin/topics/admin-cli.adoc
matthewhelmke: Linguistically, this is a more nuanced question. On the one hand, the phrase "...you access..." is a subject-verb phrase that is direct and clear, which is a hallmark of good technical documentation. On the other hand, the content is actually intending to communicate that "https:" must be used. While "you access" is correct, "you must access" is clearer.

"you have to access" is correct in terms of meaning, but "you must access" is a stronger standard in technical documentation, so I suggest using "you must access."

server_admin/topics/admin-cli.adoc
matthewhelmke: The style guide used for our documentation requires "ID" to always be capitalized, both letters. So, if we are talking about a generic "client ID," what I have written in this sentence is what we should use.

If we are talking about a specific attribute or variable, we should spell it exactly as written/required in the command, 'clientId'. We are not doing that here.

While reading this, I think we can make the section clearer. First, are these two login methods used frequently, say each about 50% of the time? Or, is one significantly more common?

If they are used with about the same frequency, we should introduce them with a phrase and put each in a bullet, like this:

There are two ways to log in using the Admin CLI:
* With a username and password - details...
* With a client ID - details...

If one is used more frequently and the other is more of a side case, I would describe the most-used option and then put the secondary option in a NOTE box.

server_admin/topics/admin-cli.adoc
matthewhelmke: The original sentence was on line 92 and read "For example, when performing an operation we specify all the information required for authentication:".

I am curious whether changing "operation" to "command" changes the meaning more than would be preferred. I think it might. Rather than fix the "an" to an "a" perhaps we change back to "operation."

We issue a command to perform an operation...

server_admin/topics/admin-cli.adoc
tanberry: Yay, a robust conversation about grammar and nouns and words! I want to join in. ;-) So on this one, I have to say that because we are not actually talking about the attribute (clientId), then we do have to capitalize 'ID'. We are talking about a noun that is a value, but in this sentence we are not talking about the attribute. We *could* of course change the sentence to talk about the attribute, and then use "clientId" (i.e. "When you log in using the clientId attribute, you only need the...").

I also suggest that we separate out the sentences about client ID, because it is smack in the middle of sentences about user name. I suggest reworking the whole paragraph to either provide the two options (username vs. client ID) as bullets, or if using client ID is the least common use case, then focus on the user name and add a Note below about the client ID. But they are mixed up too much in current paragraph, for sure.

@mstruk do we know whether most users use username or client ID?

And, sorry to say, unless you are talking about the attribute called 'username', it needs to be two separate words. ;-)

server_admin/topics/admin-cli.adoc
matthewhelmke: First off, we are not listing the full path name. There is a variable in there that stands in for /home/username. The true full path name would be /home/username/.keycloak/kcadm.config.

I suggest rephrasing to say this:

By default, the Admin CLI automatically maintains a configuration file called [filename]`kcadm.config` located at [filename]`$HOME/.keycloak/kcadm.config`. On Windows, the file is located at [filename]`%HOMEPATH%\.keycloak\kcadm.config`.

UNLESS there is a reason that the true, full path name is actually required rather than the version with the variable...in that case, we need to edit to list the full path name.

server_admin/topics/admin-cli.adoc
tanberry: @leswilliams44 where did this new sentence with "an command" come from? Was it copied/pasted from elsewhere?

server_admin/topics/admin-cli.adoc
matthewhelmke: There is a little more information required. The "with non-default permissions" is a pretty important bit of info. 😄 How about this?

If the directory already exists and its permissions are different from the {project_name} default permissions, those permissions will not be reset. In this case, it is recommended that you check and change the permissions to the {project_name} defaults.

server_admin/topics/admin-cli.adoc
matthewhelmke: That is good information! Does this express it more clearly?

If your unique circumstances require you to avoid storing secrets inside a configuration file, you can do so. It will be less convenient and you will have to make more token requests. To not store secrets, use the [command]`--no-config` option with all your commands and specify all the authentication information needed by the [command]`config credentials` command with each [command]`kcadm` invocation.

server_admin/topics/admin-cli.adoc
matthewhelmke: 👍

server_admin/topics/admin-cli.adoc
matthewhelmke: It is always tough to maintain that delicate balance between aesthetics and usefulness. The hard question that I can't yet answer for this project is whether our users will know to look in the '===Realm operations' section to find the info they need or whether they need the sign posts provided by a heading listed in the sidebar.

In the short term, I will defer to @mstruk's opinion here, since he is much more familiar with this project and I'm the newb. :)

server_admin/topics/admin-cli.adoc
matthewhelmke: 👍

server_admin/topics/admin-cli.adoc
matthewhelmke: If it is in code, everything needs to be exact.

If it is in English, then the standard in our adopted style guide is that "ID" is always capitalized.

server_admin/topics/admin-cli.adoc
matthewhelmke: 👍

server_admin/topics/admin-cli.adoc
matthewhelmke: 👍

server_admin/topics/admin-cli.adoc
matthewhelmke: 👍

server_admin/topics/admin-cli.adoc
matthewhelmke: That is clearer. Thank you for helping sort out the technical bits!

server_admin/topics/admin-cli.adoc
tanberry: I would separate this sentence into two sentences: "By default, the Admin CLI automatically maintains a configuration file called [filename]`kcadm.config` located under the user's home directory. In Linux-based systems, the full path name is [filename]`$HOME/.keycloak/kcadm.config`. On Windows, the full path name is [filename]`%HOMEPATH%\.keycloak\kcadm.config`.

However, can we really say "full" path name?? Do we even know the full path name, always?

It looks like [filename] is a variable, not sure how that will work in front of the full path name? 

Also, the phrase "automatically maintains a config file" is unclear to me. Does "maintain" mean that the config file is automatically written to and also continually updated? Or does it only meant that the config file is created automatically, but not written to? Can it be manually edited? If it is important to call out that the file is maintained (written to), and if it is important to note that the file can not be manually edited, then let's say that.

server_admin/topics/admin-cli.adoc
matthewhelmke: Is this form used anywhere?
'"name":"operations"'
If not, I agree. If so, is it introduced elsewhere/is it important?

server_admin/topics/admin-cli.adoc
matthewhelmke: 👍

matthewhelmke: 👍

matthewhelmke: The original said, "You can use `--userid` if you want to specify the user by using `id` attribute." This was clear and accurate and I think we can go back to the original phrasing.

matthewhelmke: I agree. (For convenience, this is to replace "For example, to remove from user `testuser` two roles defined on client `realm management`...", line 1069.)

matthewhelmke: Line 1080. I agree.

matthewhelmke: Line 1665. Nice catch!

matthewhelmke:
@mstruk I went through and found the line numbers that are missing in the later comments. :)

server_admin/topics/admin-cli.adoc
tanberry: Agree with @mstruk that the current phrasing makes it sound like a good idea. I would rearrange the words (and maybe make is a Note) to say:

 "If you choose to not store secrets inside a configuration file, you can use the [command]`--no-config` option with all your commands and specify all of the authentication information that is needed by the [command]`config credentials` command with each [command]`kcadm` invocation. Be aware that this approach is less convenient and you have to make more token requests.  

However,  the [command]`--no-config` option is useful for when a only single operation is needed."

server_admin/topics/admin-cli.adoc
tanberry: As good tech writers, we avoid the future tense and instead write in present tense. So I suggest "If the directory already exists with non-default permissions, those permission are not updated."

server_admin/topics/admin-cli.adoc
tanberry: agree...

tanberry (commented):
It's great to see all the edits and the edits of the edits, so I added some edits of the edits' edits. ;-) Great stuff.

server_admin/topics/admin-cli.adoc
mstruk: @tanberry @matthewhelmke The issue here is that the text in fact is not talking about a client ID as far as I understand what ID means. Client has two distinct attributes that can both be used as an ID. One is called 'id' which is a 30 chars long random string (UUID actually) and is not very practical to use. The other is called 'clientId' which is a human readable name - but a logical name, not a caption or descriptive name for display. This name does not change so it can be used as ID. We use the second one as ID in this case. Thus talking about ID rather than clientId obfuscates things. While in this particular case we don't use 'id' attribute, it is in fact used in other cases where we also refer to it as ID. Thus when we say ID we sometimes refer to 'id', and sometimes to 'clientId' attribute. That is the point I was trying to highlight here. There would be no misunderstanding if we could just use 'clientId' when we refer to 'clientId' attribute. For readability I used 'client id' in several places instead of 'clientId'. But that was apparently misleading already as it resulted in being replaced as client ID which obfuscates it even more.

server_admin/topics/admin-cli.adoc
mstruk: To me 'command' and 'operation' mean the same. Go with whatever sounds better.

server_admin/topics/admin-cli.adoc
mstruk: They both sound good to me.

server_admin/topics/admin-cli.adoc
mstruk: It's not standard practice but I did it once or twice as well. It seems leaving it there doesn't actually increase the probability of acting to address it :)
We have a build plugin that checks external links, and it's failing all the time therefore we already have (maybe many) broken links in our docs.

Maybe best if @stianst chimes in on this one.

server_admin/topics/admin-cli.adoc
mstruk: It doesn't have to say full path name. 'Located at' sounds good to me.

@tanberry Yes, 'maintained' means it is written to, and updated. As session progresses, the access token reaches it's time to live, and is automatically refreshed. The new one is saved in the file. It can be manually edited because it's a JSON file but I don't see a use case where that would be needed. We don't describe file's format, and we don't instruct users to change it directly. The important thing here is to point out that this file exists, and that it contains some secrets and so it should not be readable to other users.

server_admin/topics/admin-cli.adoc
mstruk: @matthewhelmke Default or non-default permission on the directory is not a factor. Mere existence of the directory means its permissions won't be touched. Non-default permissions is actually misleading the way I used it and should be removed.

We're not talking about Keycloak server here so this is not the right context for using {project_name}.

@tanberry What about simply 'If the directory already exists, its permissions are not updated.'

server_admin/topics/admin-cli.adoc
mstruk: I like @matthewhelmke suggestion better (sorry @tanberry :)

server_admin/topics/admin-cli.adoc
mstruk: Don't get me wrong - I know that using labels instead of subtitles is a hack. The real fix is to have a config setting that allows me to control to what depth subtitles should be included in navigation, and allow such control on a per-chapter basis. But I think we first need to find this fix before we can merge this change of labels to subtitles.

server_admin/topics/admin-cli.adoc
mstruk: What about English that refers to code which is how we use ID throughout this document?

You can't refer to 'clientId' attribute in JSON document by calling it client ID. But that's what 'client id' as used in many places actually refers to - to 'clientId' attribute in a JSON document. Imagine that we're talking about a product record structured with fields: SKU, Name, Price. Then throughout the document you refer to SKU by constantly saying ID.
Does that make any sense?

server_admin/topics/admin-cli.adoc
mstruk: No, this "name":"operations" was an error in the first place and is fixed with the previous suggestion.

server_admin/topics/admin-cli.adoc
matthewhelmke: I see. I didn't understand that until now. Perhaps we should just call this 'clientId' everywhere. What do you think?

server_admin/topics/admin-cli.adoc
matthewhelmke: 👍

server_admin/topics/admin-cli.adoc
matthewhelmke: I agree. I never want to refer to a SKU by saying ID. If that is what we are doing, we need to fix it.

Any time we refer specifically to code, we absolutely must use the exact same string as is in the code, such as 'clientId'. When we are speaking generically, we should use 'ID'. For example, we might say this: Refer to 'clientId' to determine the client ID. Okay, that is a weak sentence, but I hope it adequately illustrates the difference.

server_admin/topics/admin-cli.adoc
matthewhelmke: 👍

server_admin/topics/admin-cli.adoc
leswilliams44: I will remove the link.

server_admin/topics/admin-cli.adoc
leswilliams44: Revised to use "you must access"

server_admin/topics/admin-cli.adoc
leswilliams44: I will use clientId unless told otherwise.

server_admin/topics/admin-cli.adoc
leswilliams44: "an command" was changed back to "an operation"

server_admin/topics/admin-cli.adoc
leswilliams44: I used the two sentences suggested by @tanberry.

server_admin/topics/admin-cli.adoc
leswilliams44: Used "If the directory already exists, its permissions are not updated.'"

server_admin/topics/admin-cli.adoc
leswilliams44: Used @matthewhelmke's suggestion

server_admin/topics/admin-cli.adoc
leswilliams44: added a colon

server_admin/topics/admin-cli.adoc
leswilliams44: @mstruk Is this a possible solution?  https://stackoverflow.com/questions/34481638/how-to-use-tocify-with-asciidoctor-for-a-dynamic-toc

leswilliams44: Here is another option.

If the title text should not be numbered or appear in the table of contents, use a heading at the appropriate level, and then set the discrete attribute above the heading.

[discrete]
==== Delete a realm
Text text text...

mstruk: Fantastic, that's it!
I tried it and it works. Do you want to add that to all fourth level subtitles in admin-cli.adoc?

matthewhelmke: 👍

matthewhelmke: @stianst , who do we need to talk to to implement this solution that @leswilliams44 found above and @mstruk has tested? The solution is being suggested because this is a very long page that has too many entries in the table of contents. The ultimate solution would be for me to break this up into multiple pages, but that isn't something I have bandwidth for today (but it is on my TODO list for the future).

https://stackoverflow.com/questions/34481638/how-to-use-tocify-with-asciidoctor-for-a-dynamic-toc

Essentially, it involves adding a js file (full source in the link above) and then running asciidoctor with different options to build.

So, it looks like a great solution for asciidoctor, but I don't know how or even if this is feasible with our current toolchain. If so, I would like us to consider it. If not, I wonder if there is another option.

Do you know, and if not, do you know who I should be asking? Thanks.

server_admin/topics/admin-cli.adoc
leswilliams44: Revised to say "You append a realm name to a collection URI to get an individual realm."

leswilliams44:
Stopped at line 525. Will continue to address all comments.

server_admin/topics/admin-cli.adoc
mstruk: @matthewhelmke The [discrete] solution doesn't require anything extra in the build. It is supported out of the box. All that has to be done is to insert [discrete]  in front of all fourth level subtitles in admin-cli.adoc.

matthewhelmke: @mstruk OH! I totally misread the earlier comment. Yes! Let's just use [discrete]. That is elegant and simple. Thank you for correcting my understanding.

leswilliams44: I will apply the [discrete] attribute to all level four headings.

server_admin/topics/admin-cli.adoc
leswilliams44: I will go through the entire document and correct the revisions to clientID to the best of my understanding. Please check my work.

server_admin/topics/admin-cli.adoc
leswilliams44: Corrected typo

server_admin/topics/admin-cli.adoc
leswilliams44: The sentence has been deleted.

server_admin/topics/admin-cli.adoc
leswilliams44: The sentence has been deleted.

server_admin/topics/admin-cli.adoc
leswilliams44: The suggested change has been made.

server_admin/topics/admin-cli.adoc
leswilliams44: The sentence has been deleted.

server_admin/topics/admin-cli.adoc
leswilliams44: The suggested change has been implemented.

leswilliams44: @matthewhelmke's suggestion was implemented.

leswilliams44: The change has been made.

leswilliams44: The change has been implemented.

leswilliams44: done

leswilliams44: typo corrected

PR #259, opened 2017-12-04
Minor typo

secondsun:
I can only make so many jokes about the typo, but I did double check that requested_toke_type is not a parameter.


matthewhelmke:
Hahaha! Thank you very much.

PR #262, opened 2017-12-05
KEYCLOAK-5734

stianst:


stianst:
Update to code https://github.com/keycloak/keycloak/pull/4797

server_admin/topics/users/user-registration.adoc
```diff
@@ -3,7 +3,12 @@
 === User Registration
 
 You can enable {project_name} to allow user self registration.  When enabled, the login page has a registration
-link the user can click on to create their new account.  Enabling registration is pretty simple.  Go to the
+link the user can click on to create their new account.
+
+When user self registration is enabled it is possible to use the registration form to detect valid usernames and emails.
+To prevent this enable <<_recaptcha,reCAPTCHA Support>>.
+
```
matthewhelmke, line 10: Are we trying to prevent them from using the registration form to detect valid usernames and emails?

If so, this is clear. If not, I am unsure what we are preventing by enabling the recaptcha support.

stianst:
Yes. We have tried elsewhere to prevent detecting valid usernames/emails. For example the login screen will display "invalid username/password" rather than just "invalid username".

If the registration form doesn't have captcha enabled an attacker could use the registration form to create lists of valid usernames/passwords. However, if the captcha is enabled that becomes rather difficult as it would have to be done manually which has much less value to an attacker.

matthewhelmke:
Okay, that makes sense and was what I was thinking when I read the addition (but wanted to confirm).

So, LGTM. 👍

PR #263, opened 2017-12-06
RHSSO-1197 for 7.1 release

matthewhelmke:
These are changes for 7.1, as requested in https://issues.jboss.org/browse/RHSSO-1197?focusedCommentId=13497734&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13497734

### DIFF
--- a/securing_apps/topics/oidc/java/jboss-adapter.adoc
+++ b/securing_apps/topics/oidc/java/jboss-adapter.adoc
@@ -103,13 +103,13 @@ $ sudo subscription-manager repos --enable=jb-eap-7-for-rhel-<RHEL_VERSION>-serv
 Install the EAP 7 adapters for OIDC using the following command:
 
 ----
-$ sudo yum install eap7-keycloak-adapter
+$ sudo yum install keycloak-adapter
 ----
 
 Install the EAP 7 adapters for SAML using the following command:
 
 ----
-$ sudo yum install eap7-keycloak-saml-adapter
+$ sudo yum install keycloak-saml-adapter
 ----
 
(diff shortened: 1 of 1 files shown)


stianst:
I'm not quite following this PR. We should document steps to install 7.1 adapters in 7.2 documentation.

matthewhelmke:
Okay. I guess I misunderstood the issue. Since there are different filenames for 7.1 and 7.2, I was thinking we would want to merge just the 7.1 change in this PR now and get it merged all the way into the existing 7.1 docs, then do a second PR just for 7.2. Since our documentation is versioned for the product, my thinking was that each release's docs only needed the info specific to that release.

Would you prefer that I add the 7.1 info into the 7.2 doc PR? If so, we can close this PR and I'll update the 7.2. Before doing so, I should ask though, how will 7.1 users know the filenames?

stianst:
7.1 documentation is managed to a different branch. In 7.1 documentation we've not documented how to install adapter RPMs at all. As it's not to long to 7.2 is out we should just not update the 7.1 docs at all. It's a bit of an effort to backport this to 7.1 docs and publish the update, which is time we could better spend on other things. Once 7.2 is out I can walk you through how we'll manage documentation for the currently released version as this process has changed quite a bit since 7.1.

For 7.2 documentation (master) it should only mention the 7.2 adapter RPMs. As we didn't document adapter RPMs for 7.1 I don't think we need to mention that the names have changed. IMO the filenames for 7.1 should not be mentioned at all in 7.2 docs.

matthewhelmke:
Okay, that makes sense. I'll close this PR. The other PR only mentions the 7.2 specific details, so as long as those details are approved as correct in review, we should be set.

PR #265, opened 2017-12-06
KEYCLOAK-5915 Support for sticky sessions managed by loadbalancer

mposolda:


server_installation/topics/clustering/booting.adoc
```diff
@@ -16,3 +16,6 @@ $ bin/domain.sh --host-config=host-master.xml
 $ bin/domain.sh --host-config=host-slave.xml
 ----
 
+It may be needed to use some additional parameters or system properties. For example parameter `-b` for the binding host. Or system property
+`jboss.node.name` to specify the name of the route as described in  <<sticky-sessions,Sticky Sessions >> section
+
```
matthewhelmke: You may need to use additional parameters or system properties. For example, the parameter `-b` for the binding host or the system property `jboss.node.name` to specify the name of the route, as described in  <<sticky-sessions,Sticky Sessions >> section

server_installation/topics/clustering/sticky-sessions.adoc
```diff
@@ -31,11 +31,39 @@ and hence Infinispan can lookup this session locally. After authentication is fi
...
 
-Some loadbalancers can be configured to add the route information by themselves instead of rely on the backend {project_name} node. However adding the route by
-the {project_name} is more clever as {project_name} knows, who is the owner of particular session entity and can route it to that node, which is not necessarily the local node.
-We plan to support the setup where the route info won't be added to AUTH_SESSION_ID cookie by the {project_name} server, but instead by the load balancer. However
-adding the cookie by {project_name} would be still the preferred option.
+The recommended step on {project_name} side is to use the system property `jboss.node.name` during startup with the value corresponding to the name of your route.
+For example `-Djboss.node.name=node1` will use the `node1` as the identification of the route. This route will be used by Infinispan caches and will be
+ attached to the AUTH_SESSION_ID cookie in case that our node is the owner of the particular key. The example of the startup command with this system property is below in the
+ <<_example-setup-with-mod-cluster,Mod Cluster Example>> section.
+
```
matthewhelmke: It is recommended on the {project_name} side to use the system property `jboss.node.name` during startup, with the value corresponding to the name of your route. For example, `-Djboss.node.name=node1` will use `node1` to identify the route. This route will be used by Infinispan caches and will be attached to the AUTH_SESSION_ID cookie when the node is the owner of the particular key. An example of the start up command with this system property can be seen in <<_example-setup-with-mod-cluster,Mod Cluster Example>>.

server_installation/topics/clustering/sticky-sessions.adoc
```diff
@@ -31,11 +31,39 @@ and hence Infinispan can lookup this session locally. After authentication is fi
...
+==== Disable adding the route
+
+Some loadbalancers can be configured to add the route information by themselves instead of rely on the backend {project_name} node. However as described above,
+adding the route by the {project_name} is recommended especially due the better performance as {project_name} knows, who is the owner of particular session entity and
+can route it to that node, which is not necessarily the local node.
+
+Still if you want to disable adding route information to the AUTH_SESSION_ID cookie by {project_name}, you can do it by adding this into your
+`RHSSO_HOME/standalone/configuration/standalone-ha.xml` file into the {project_name} subsystem configuration:
+
```
matthewhelmke: Typically, the route name be the same name as your back end host, but it is not necessary. You can use a different route name, for example if you want to hide the host name of your {project_name} server inside your private network.

==== Disable adding the route

Some load balancers can be configured to add the route information by themselves instead of relying on the back end {project_name} node. However, as described above, adding the route by the {project_name} is recommended. This is because when done this way performance improves, since {project_name} is aware of the entity that is the owner of particular session and can route to that node, which is not necessarily the local node.

You are permitted to disable adding route information to the AUTH_SESSION_ID cookie by {project_name}, if you prefer, by adding the following into your `RHSSO_HOME/standalone/configuration/standalone-ha.xml` file in the {project_name} subsystem configuration:

matthewhelmke (changes requested):
I have suggested a few changes, mostly to make the language clearer. When incorporated, this is ready.

mposolda:
@matthewhelmke Changes incorporated. Thanks! I hope my English improve one day... ;)

matthewhelmke (approved):

matthewhelmke:
Looks good to me, @mposolda. English is hard, even for native speakers. Yours is better than you think. Keep up the good work, and thanks!

mposolda:
@matthewhelmke Thanks!

PR #267, opened 2017-12-14
KEYCLOAK-2944 Document how to restrict admin endpoints to separate po…

stianst:
…rt or to range of IP addresses


server_admin/topics/threat/admin.adoc
```diff
@@ -0,0 +1,95 @@
+
+=== Admin Endpoints and Console
+
+{project_name} administrative REST API and web console are by default exposed on the same port as non-admin usage. If you
+are exposing {project_name} on the Internet we recommend not exposing the admin endpoints on the Internet as well.
+
```
matthewhelmke: The {project_name} administrative REST API and the web console are exposed by default on the same port as non-admin usage. If you are exposing {project_name} on the Internet, we recommend not also exposing the admin endpoints on the Internet.

matthewhelmke (commented):
Looks good to me. I have one suggested change for the wording of the introductory paragraph.

stianst:
Suggestion added

PR #269, opened 2017-12-14
Fix typos in Upgrading

k-tamura:
There are several typos in the Upgrading guide.


matthewhelmke:
Thank you. Your help is appreciated!

PR #291, opened 2018-01-16
remove title modifier for paypal section

sebastienblanc:
This was breaking the left side TOC. 
Further, I think we should review the subsection inside each social provider doc using the `.subsection` doesn't seem to render correctly


matthewhelmke (approved):
Thank you for fixing this!

I took a look and see what you mean by the rendering. I wonder if the CSS for rendering the `.subsection` is too similar to regular text (or perhaps doesn't even exist?).

matthewhelmke:
Adding a screenshot of the problem, for reference. "Add identity provider" should be styled differently than the regular text in the paragraph above it.

screenshot from 2018-01-16 09-03-37

PR #293, opened 2018-01-18
Remove spurious extra heading, and menu item, from PayPal section

nicdoye:
The documentation contains a section "12.4.8. Do the following changes:" which should really just be part of the PayPal section (12.4.7).


matthewhelmke:
Thank you. We recently approved, but have not yet merged another PR with the same fix, #291 .

I'll close this, but only for that reason. Your contribution and attention to detail are very much appreciated.

PR #294, opened 2018-01-18
KEYCLOAK-6102 fixes from QE review

matthewhelmke:
Here are the comments from the report, with my responses. @stianst will you please select a knowledgeable person to help me answer the parts I don't have the knowledge to answer?

> I've done the Client Registration CLI review and I've found several issues.
> [RH-SSO only] Many places throughout the whole Chapter 6. of Securing Applications and Services Guide
> Keycloak instead of RH-SSO is mentioned everywhere

I have changed this to `{project_name}` in text, but not in code. I need someone who knows the code to help me understand where/if this should be changed in code examples.

> Chapter 6.2. Configuring a client for use with the Client Registration CLI
> [RH-SSO only] In the step 6, instead of the word “Save” there’s some weird and unexpected SAVE button.

Fixed.

> The  whole chapter is a bit erratic and incomplete. Doesn’t e.g. mention that Standard Flow should be disabled (so no valid redirect URLs would be required) or that Direct Access Grants needs to be enabled to login as a user when not using service accounts. Service account roles are not mentioned either.

I need further guidance and information on the technical aspects in order to write up the missing information.

> 6.4.3. Initial Access and Registration Access Tokens
> The first sentence of the first paragraph doesn’t make sense.

I have edited the entire paragraph.

### DIFF
--- a/securing_apps/topics/client-registration/client-registration-cli.adoc
+++ b/securing_apps/topics/client-registration/client-registration-cli.adoc
@@ -166,7 +166,7 @@ You might want to avoid storing secrets inside a configuration file by using the
 [[_initial_access_and_registration_access_tokens]]
 ==== Initial Access and Registration Access Tokens
 
-Developers who do not have an account configured at the {project_name} server they want to use can instead use the Client Registration CLI. This is possible only when the realm administrator issues a developer an Initial Access Token. It is up to the realm administrator to decide how and when to issue and distribute these tokens. The realm administrator can limit the maximum age of the Initial Access Token and the total number of clients that can be created with it.
+Developers who do not have an account configured at the {project_name} server they want to use can use the Client Registration CLI. This is possible only when the realm administrator issues a developer an Initial Access Token. It is up to the realm administrator to decide how and when to issue and distribute these tokens. The realm administrator can limit the maximum age of the Initial Access Token and the total number of clients that can be created with it.
 
 Once a developer has an Initial Access Token, the developer can use it to create new clients without authenticating with [command]`kcreg config credentials`. The Initial Access Token can be stored in the configuration file or specified as part of the [command]`kcreg create` command.
 
--- a/securing_apps/topics/client-registration/client-registration-cli.adoc
+++ b/securing_apps/topics/client-registration/client-registration-cli.adoc
@@ -39,7 +39,7 @@ By default, the server recognizes the Client Registration CLI as the [filename]`
 . Create a new client (for example, [filename]`reg-cli`) if you want to use a separate client configuration for the Client Registration CLI.
 . Specify which [filename]`clientId` to use (for example, [command]`--client reg-cli`) when running [command]`kcreg config credentials`.
 . Enable service accounts if you want to use a service account associated with the client by selecting a client to edit in the *Clients* section of the `Admin Console`.
-. Under *Settings*, change the *Access Type* to *Confidential*, toggle the *Service Accounts Enabled* setting to *On*, and click [btn]`Save`.
+. Under *Settings*, change the *Access Type* to *Confidential*, toggle the *Service Accounts Enabled* setting to *On*, and click *Save*.
 +
 [NOTE]
 ====
@@ -51,9 +51,9 @@ You can configure either [filename]`Client Id and Secret` or [filename]`Signed J
 [[_installing_client_registration_cli]]
 === Installing the Client Registration CLI
 
-The Client Registration CLI is packaged inside the Keycloak Server distribution. You can find execution scripts inside the [filename]`bin` directory. The Linux script is called [filename]`kcreg.sh`, and the Windows script is called [filename]`kcreg.bat`.
+The Client Registration CLI is packaged inside the {project_name} Server distribution. You can find execution scripts inside the [filename]`bin` directory. The Linux script is called [filename]`kcreg.sh`, and the Windows script is called [filename]`kcreg.bat`.
 
-Add the Keycloak server directory to your [filename]`PATH` when setting up the client for use from any location on the file system.
+Add the {project_name} server directory to your [filename]`PATH` when setting up the client for use from any location on the file system.
 
 For example, on:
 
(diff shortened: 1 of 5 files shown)


securing_apps/topics/client-registration/client-registration-cli.adoc
```diff
@@ -166,7 +166,7 @@ You might want to avoid storing secrets inside a configuration file by using the
 [[_initial_access_and_registration_access_tokens]]
 ==== Initial Access and Registration Access Tokens
 
-Developers who do not have an account configured at the Keycloak server they want to use can use the Client Registration CLI. That is possible when the realm administrator issues a developer an Initial Access Token. It is up to the realm administrator to decide how to issue and distribute these tokens. The realm administrator can limit the maximum age of the Initial Access Token and the total number of clients that can be created with it.
+Developers who do not have an account configured at the {project_name} server they want to use can instead use the Client Registration CLI. This is possible only when the realm administrator issues a developer an Initial Access Token. It is up to the realm administrator to decide how and when to issue and distribute these tokens. The realm administrator can limit the maximum age of the Initial Access Token and the total number of clients that can be created with it.
```
mstruk: The idea here is that you don't need to have an account in order to use Client Registration CLI. You can simple be given an Initial Access Token and authenticate with that.

So 'instead' would not be the tight word here, rather we can say 'also'.

mstruk (changes requested):

securing_apps/topics/client-registration/client-registration-cli.adoc
matthewhelmke: I am happy to make that change. I have more comments coming in from QE, so I plan to push another commit tomorrow morning which will contain this change and those.

EDIT: after some thought, I decided to delete 'instead' and not replace it with anything. I think '...they want to use can use the...' is sufficiently clear and doesn't accidentally imply anything we don't want to say.

matthewhelmke:
The new commit fixes the issue brought up by @mstruk and also the following from QE:

> Whole document
> 
>     namespace versions
>         <server xmlns="urn:jboss:domain:1.4"> -> <server xmlns="urn:jboss:domain:5.0">
>         <subsystem xmlns="urn:jboss:domain:security:1.2"> -> <subsystem xmlns="urn:jboss:domain:security:2.0">
> 
>     rh-sso version
>         3.4.2.Final-redhat1 -> 3.4.3.Final-redhat-2
> 
> 2.1.2.1. Installing the adapter
> 
>     2x : Any other server but Wildfly 11 - not sure whether there should be Wildlfy
> 
> 3.1.2.1. Adapter Installation
> 
>     Any other server but Wildfly 11 - same as above
> 
> Chapter 4. Docker Registry Configuration
> 
>     corrupted link: ...server_installation_and_configuration_guide/#_profiles -> server_installation_and_configuration_guide/#profiles
> 
> Chapter 7. Token Exchange
> 
>     corrupted link: ...server_installation_and_configuration_guide/#_profiles -> server_installation_and_configuration_guide/#profiles

topics/templates/document-attributes-product.adoc
```diff
@@ -2,8 +2,8 @@
 :project_community: false
 :project_product: true
 :project_version: 7.2.0.GA
-:project_versionMvn: 3.4.2.Final-redhat1
-:project_versionNpm: 3.4.2.Final-redhat1
+:project_versionMvn: 3.4.2.Final-redhat-2
```
vramik: Hey @matthewhelmke, the version should be 3.4.3.Final-redhat-2, not 3.4.2.Final-redhat-2

vmuzikar:
@matthewhelmke Proposed changes to the mentioned "Chapter 6.2. Configuring a client for use with the Client Registration CLI" (was not sure how to create a link to chapter 6.1):
```text
. Create a new client (for example, [filename]`reg-cli`) if you want to use a separate client configuration for the Client Registration CLI.
. Toggle the *Standard Flow Enabled* setting it to *Off*.
. Strengthen the security by configuring the client [filename]`Access Type` as [filename]`Confidential` and selecting *Credentials > ClientId and Secret*.
[NOTE]
====
You can configure either [filename]`Client Id and Secret` or [filename]`Signed JWT` under the *Credentials* tab .
====
. Enable service accounts if you want to use a service account associated with the client by selecting a client to edit in the *Clients* section of the `Admin Console`.
.. Under *Settings*, change the *Access Type* to *Confidential*, toggle the *Service Accounts Enabled* setting to *On*, and click *Save*.
.. Click *Service Account Roles* and select desired roles to configure the access for the service account. For the details on what roles to select, see "6.1. Configuring a new regular user for use with Client Registration CLI".
. Toggle the *Direct Access Grants Enabled* setting it to *On* if you want to use a regular user account instead of a service account.
. If the client is configured as [filename]`Confidential`, provide the configured secret when running [command]`kcreg config credentials` by using the [command]`--secret` option.
. Specify which [filename]`clientId` to use (for example, [command]`--client reg-cli`) when running [command]`kcreg config credentials`.
. With the service account enabled, you can omit specifying the user when running [command]`kcreg config credentials` and only provide the client secret or keystore information.
```
@mstruk What do you think? Is that correct?

mstruk:
@vmuzikar Is there a diff, to see the changes?
In any case, I couldn't really say without trying out the described steps. Would you like me to try out these instructions, and give feedback?

vmuzikar:
@mstruk I've reordered some steps (to be more logical) and added a few more as well (turning off Standard Flow, enabling Direct Access Grant, ...). So the diff would imho be too big and confusing.

topics/templates/document-attributes-product.adoc
matthewhelmke: Thank you.

matthewhelmke:
@vmuzikar and @mstruk , I think the suggested changes would be better in a separate PR. We would then have a diff available and it would be much easier to evaluate the proposed changes.

vmuzikar:
@matthewhelmke Agree. I'll open a new PR with my suggested changes once this PR is final and merged (to avoid conflicts).

stianst:
Fixing and splitting into two separate PRs

stianst:
https://github.com/keycloak/keycloak-documentation/pull/311

stianst:
https://github.com/keycloak/keycloak-documentation/pull/313

PR #295, opened 2018-01-22
KEYCLOAK-6286 Added docs for OIDC Compatibility Modes to the Upgrade …

mposolda:
…guide

### DIFF
--- a/upgrading/topics/keycloak/changes.adoc
+++ b/upgrading/topics/keycloak/changes.adoc
@@ -1,5 +1,22 @@
 == Migration Changes
 
+=== Migrating to 3.4.2
+
+==== Added session_state parameter to OpenID Connect Authentication Response
+
+The OpenID Connect Session Management specification requires that the parameter `session_state` is present in the OpenID Connect Authentication Response.
+
+In past releases, we did not have this parameter, but now {project_name} adds this parameter by default, as required by the specification.
+
+However, some OpenID Connect / OAuth2 adapters, and especially older {project_name} adapters, may have issues with this new parameter.
+
+For example, the parameter will be always present in the browser URL after successful authentication to the client application.
+In these cases, it may be useful to disable adding the `session_state` parameter to the authentication response. This can be done
+for the particular client in the {project_name} admin console, in client details in the section with `OpenID Connect Compatibility Modes`,
+described in <<_compatibility_with_older_adapters>>. There is the `Exclude Session State From Authentication Response` switch,
+which can be turned on to prevent adding the `session_state` parameter to the Authentication Response.
+
+
 === Migrating to 3.2.0
 
 ==== New Password Hashing algorithms
--- a/upgrading/topics/rhsso/changes-72.adoc
+++ b/upgrading/topics/rhsso/changes-72.adoc
@@ -32,3 +32,17 @@ Microsoft JDBC Driver 6.0 requires additional dependency added to the JDBC drive
 ----
 <module name="javax.xml.bind.api"/>
 ----
+
+=== Added session_state parameter to OpenID Connect Authentication Response
+
+The OpenID Connect Session Management specification requires that the parameter `session_state` is present in the OpenID Connect Authentication Response.
+
+In RH-SSO 7.1, we did not have this parameter, but now {project_name} adds this parameter by default, as required by the specification.
+
+However, some OpenID Connect / OAuth2 adapters, and especially older {project_name} adapters (such as RH-SSO 7.1 and older), may have issues with this new parameter.
+
+For example, the parameter will be always present in the browser URL after successful authentication to the client application.
+If you use RH-SSO 7.1 or a legacy OAuth2 / OpenID Connect adapter, it may be useful to disable adding the `session_state` parameter to the authentication response.
+This can be done for the particular client in the {project_name} admin console, in client details in the section with `OpenID Connect Compatibility Modes`,
+described in <<_compatibility_with_older_adapters>>. There is the `Exclude Session State From Authentication Response` switch,
+which can be turned on to prevent adding the `session_state` parameter to the Authentication Response.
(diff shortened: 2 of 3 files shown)


mposolda:
Implementation PR: https://github.com/keycloak/keycloak/pull/4928

stianst:
We also need to add the migration note to RH-SSO 7.2 changes. It's in rhsso/changes-72.adoc

mposolda:
@stianst @matthewhelmke Added section about RH-SSO 7.2 changes as well.

FYI. The Travis failure doesn't seem to be caused by this PR, but by the KEYCLOAK-6353 . I've sent another PR to fix that: https://github.com/keycloak/keycloak-documentation/pull/298 . After fixing that, the travis re-run should hopefully pass.

upgrading/topics/keycloak/changes.adoc
```diff
@@ -1,5 +1,20 @@
...
+OpenID Connect Session Management specification requires that parameter `session_state` is present in the OpenID Connect Authentication Response.
+In the older versions, we didn't add this parameter, but now {project_name} adds this parameter by default as required per the specification.
+
+Problem is, that some OpenID Connect / OAuth2 adapters and especially older {project_name} adapters may have issues with this new parameter.
+For example parameter will be always present in the browser URL after successful authentication to the client application. For those cases,
+it may be useful to disable adding session_state parameter to the authentication response. It can be done for the particular client in
+{project_name} admin console in client details and section with `OpenID Connect Compatibility Modes` described in <<_compatibility_with_older_adapters>> .
+There is the switch `Exclude Session State From Authentication Response`, which can be turned on to not add the `session_state` parameter
+to the Authentication Response.
+
```
matthewhelmke: The OpenID Connect Session Management specification requires that the parameter `session_state` is present in the OpenID Connect Authentication Response.

In past releases, we did not have this parameter, but now {project_name} adds this parameter by default, as required by the specification.

However, some OpenID Connect / OAuth2 adapters, and especially older {project_name} adapters, may have issues with this new parameter.

For example, the parameter will be always present in the browser URL after successful authentication to the client application. In these cases, it may be useful to disable adding the `session_state` parameter to the authentication response. This can be done for the particular client in the {project_name} admin console, in client details in the section with `OpenID Connect Compatibility Modes`, described in <<_compatibility_with_older_adapters>>. There is the `Exclude Session State From Authentication Response` switch, which can be turned on to prevent adding the `session_state` parameter to the Authentication Response.

upgrading/topics/rhsso/changes-72.adoc
```diff
@@ -31,4 +31,16 @@ Microsoft JDBC Driver 6.0 requires additional dependency added to the JDBC drive
...
+
+OpenID Connect Session Management specification requires that parameter `session_state` is present in the OpenID Connect Authentication Response.
+In the RH-SSO 7.1, we didn't add this parameter, but now {project_name} 7.2 server adds this parameter by default as required per the specification.
+
+Problem is, that some OpenID Connect / OAuth2 adapters and especially older {project_name} adapters (RH-SSO 7.1 and older) may have issues with this new parameter.
+For example parameter will be always present in the browser URL after successful authentication to the client application. So if you use RH-SSO 7.1 or
+some other legacy OAuth2 / OpenID Connect adapter, it may be useful to disable adding session_state parameter to the authentication response.
+It can be done for the particular client in {project_name} admin console in client details and section with `OpenID Connect Compatibility Modes`
+described in <<_compatibility_with_older_adapters>> . There is the switch `Exclude Session State From Authentication Response`, which can be
+turned on to not add the `session_state` parameter to the Authentication Response.
```
matthewhelmke: The OpenID Connect Session Management specification requires that the parameter `session_state` is present in the OpenID Connect Authentication Response.

In RH-SSO 7.1, we did not have this parameter, but now {project_name} adds this parameter by default, as required by the specification.

However, some OpenID Connect / OAuth2 adapters, and especially older {project_name} adapters (such as RH-SSO 7.1 and older), may have issues with this new parameter.

For example, the parameter will be always present in the browser URL after successful authentication to the client application. If you use RH-SSO 7.1 or a legacy OAuth2 / OpenID Connect adapter, it may be useful to disable adding the `session_state` parameter to the authentication response. This can be done for the particular client in the {project_name} admin console, in client details in the section with `OpenID Connect Compatibility Modes`, described in <<_compatibility_with_older_adapters>>. There is the `Exclude Session State From Authentication Response` switch, which can be turned on to prevent adding the `session_state` parameter to the Authentication Response.

upgrading/topics/upgrade_adapters.adoc
```diff
@@ -4,6 +4,18 @@ It is important that you upgrade {project_name} server first, and then upgrade t
...
+
+As mentioned above, we try to support the newer version of {project_name} server to work with older version of the adapters.
+However in some cases, we need to do fixes on the {project_name} server side, which may break compatibility with older version
+of the adapters. For example when we implement some new aspects of the OpenID Connect specification, which older client adapter
+is not aware off.
+
+For those cases, we added Compatibility modes. For OpenId Connect clients, there is the section named `OpenID Connect Compatibility Modes` in
+the {project_name} admin console on the page with client details. Here you can disable some new aspects of {project_name} server
+to preserve compatibility with older client adapters. More details are available in the tooltips of individual switches.
+
```
matthewhelmke: As mentioned above, we try to support newer release versions of {project_name} server working with older release versions of the adapters. However, in some cases we need to include fixes on the {project_name} server side which may break compatibility with older versions of the adapters. For example, when we implement new aspects of the OpenID Connect specification, which older client adapter versions were not aware of.

In those cases, we added Compatibility modes. For OpenId Connect clients, there is a section named `OpenID Connect Compatibility Modes` in the {project_name} admin console, on the page with client details. Here, you can disable some new aspects of the {project_name} server to preserve compatibility with older client adapters. More details are available in the tool tips of individual switches.

matthewhelmke (changes requested):
Looks good. My language edits are light.

mposolda:
@matthewhelmke Refactored based on your suggestions. Thanks a lot!

matthewhelmke (approved):
Looks good. Thank you!

mposolda:
@matthewhelmke Thanks!

PR #296, opened 2018-01-23
Fix typos in Authorization Services

k-tamura:
I found several typos in Authorization Services too.


matthewhelmke (approved):
Looks good to me. Thank you!

PR #302, opened 2018-01-24
KEYCLOAK-6118 added upgrade via RPM to guide

matthewhelmke:
I need to confirm I captured this information correctly before merging.

### DIFF
--- a/upgrading/topics/install_new_version.adoc
+++ b/upgrading/topics/install_new_version.adoc
@@ -4,9 +4,23 @@
 
 It is important that you upgrade {project_name} server before upgrading the adapters.
 
-To upgrade {project_name} server, complete the following steps:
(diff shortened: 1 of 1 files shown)


pdrozd:
@matthewhelmke it seems that my comment wans not accurate. It is about chapter "Upgrading the EAP Adapter". 
We need to split the chapter for micro and minor release. For zips there is same procedure for micro and minor: Download, remove dir...
For the rpms we need micro upgrade where is sufficient to do "yum update" and minor upgrade:
1. remove the old adapter by yum remove
2. install the new adapter yum install

matthewhelmke:
@pdrozd , okay. In that case, I will close this PR and submit a new one soon once I have rewritten that content.

PR #298, opened 2018-01-24
KEYCLOAK-6353 Tests broken in documentation due the invalid redirect

mposolda:

PR #304, opened 2018-01-24
Fix link to token exchange spec

stianst:


matthewhelmke (approved):
Looks good to me. If we want perfection, the link ends up redirecting from HTTP to HTTPS, so we could also add the s to both links.

mposolda:
@stianst @matthewhelmke I've actually did the same thing in the morning, just added also 's' to the links to avoid redirecting to HTTPS as Matthew mentioned. Looks we can close this?

mposolda:
Jira is: https://issues.jboss.org/browse/KEYCLOAK-6353

matthewhelmke:
Yes, @mposolda did this in #298, which I already merged. Closing this issue.

PR #306, opened 2018-01-25
Fix typos in Server Admin

k-tamura:


matthewhelmke (approved):
Thank you. All these fixes are appreciated greatly!

PR #308, opened 2018-01-26
Fix typos in Server Installation

k-tamura:


matthewhelmke:
Thank you. I appreciate you catching and fixing these! We will merge them soon.

matthewhelmke (approved):

PR #317, opened 2018-02-02
Fix typos in Server Development

k-tamura:


matthewhelmke (approved):
Thank you for helping with these!

PR #319, opened 2018-02-07
KEYCLOAK-6289 Documentation for theme selector SPI

stianst:


server_development/topics/themes-selector.adoc
```diff
@@ -0,0 +1,12 @@
+= Theme Selector
+
+By default the theme configured for the realm is used, with the exception of clients being able to override the login
+theme. This behavior can be changed through the Theme Selector SPI.
+
+This could for example be used to create to select different themes for desktop and mobile devices by looking at the
+user agent header.
+
```
matthewhelmke: This could be used to create to select different themes for desktop and mobile devices by looking at the user agent header, for example.

matthewhelmke (dismissed):
One minor suggestion, but not a deal-breaker.

stianst:
Updated

matthewhelmke (dismissed):

stianst:
@matthewhelmke tests failed as the providers chapter wasn't available in prod docs. I've rewritten the providers chapter (quickly, it can do with more work in the future) to allow showing it in prod docs.

matthewhelmke (approved):
@stianst Thank you.

stianst:
@matthewhelmke you probably want to add the providers chapter to list of stuff to improve as it's not very nice (I wrote it so I'm allowed to say it ;))

matthewhelmke:
@stianst I'll add it to my list! 😄

PR #328, opened 2018-02-13
Update Kerberos CLI Documentation

slominskir:
The old user-federation/instances command was replaced with the new components command.  Fields updated as arrays are now used and single quotes are required around some attributes.  The parentId attribute was added.

### DIFF
--- a/server_admin/topics/admin-cli.adoc
+++ b/server_admin/topics/admin-cli.adoc
@@ -1471,14 +1471,14 @@ $ kcadm.sh create identity-provider/instances -r demorealm -s alias=stackoverflo
 [discrete]
 ==== Configuring a Kerberos storage provider
 
-. Use the [command]`create` command against the [filename]`user-federation/instances` endpoint.
-. Specify [command]`kerberos` as a value of the [command]`providerName` attribute.
-+
-For example:
+. Use the [command]`create` command against the [filename]`components` endpoint.
+. Specify realm id as a value of the [command]`parentId` attribute.
+. Specify [command]`kerberos` as the value of the [command]`providerId` attribute, and [command]`org.keycloak.storage.UserStorageProvider` as the value of the [command]`providerType` attribute.
+. For example:
 +
 [options="nowrap"]
 ----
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
Thank you!

server_admin/topics/admin-cli.adoc
```diff
@@ -1471,14 +1471,14 @@ $ kcadm.sh create identity-provider/instances -r demorealm -s alias=stackoverflo
...
+. For example:
 +
 [options="nowrap"]
 ----
-$ kcadm.sh create user-federation/instances -r demorealm -s providerName=kerberos -s priority=0 -s config.debug=false -s config.allowPasswordAuthentication=true -s 'config.editMode="UNSYNCED"' -s config.updateProfileFirstLogin=true -s config.allowKerberosAuthentication=true -s 'config.kerberosRealm="KEYCLOAK.ORG"' -s 'config.keyTab="http.keytab"' -s 'config.serverPrincipal="HTTP/localhost@KEYCLOAK.ORG"'
+$ kcadm.sh create components -r demorealm -s parentId=demorealmId -s id=demokerberos -s name=demokerberos -s providerId=kerberos -s providerType=org.keycloak.storage.UserStorageProvider -s 'config.priority=["0"]' -s 'config.debug=["false"]' -s 'config.allowPasswordAuthentication=["true"]' -s 'config.editMode=["UNSYNCED"]' -s 'config.updateProfileFirstLogin=["true"]' -s 'config.allowKerberosAuthentication=["true"]' -s 'config.kerberosRealm=["KEYCLOAK.ORG"]' -s 'config.keyTab=["http.keytab"]' -s 'config.serverPrincipal=["HTTP/localhost@KEYCLOAK.ORG"]' -s 'config.cachePolicy=["DEFAULT"]'
```
mstruk, line 1481: Rather than: -s parentId=demorealmId -s id=demokerberos
it should be: -s parentId=demorealm

Attribute 'id' should not be set - it is silently ignored if it is set. And parentId should be the same as the value of -r parameter (in this case 'demorealm').

mstruk (changes requested):
There's a tiny change needed but other than that it looks good. Thanks for the fix.

slominskir:
@mstruk - I'm not sure I follow.   I've observed that the id is not ignored, it is used.  If no ID is specified a randomly generated one is used.  Also, I used the placeholder demorealmId instead of demorealm to indicate that the name and id are in fact two separate and independent fields and do not have to match.  It turns out that making them have identical values is allowed and perhaps makes things easier, but they can be different.

stianst:
@mposolda can you review this please?

mposolda:
@slominskir Looks good. Could you please create JIRA? Or if you already created one, link it there and this PR from the JIRA?

slominskir:
https://issues.jboss.org/browse/KEYCLOAK-6583

mposolda (approved):

mposolda:
@slominskir Thanks!

PR #334, opened 2018-03-01
[KEYCLOAK-6116] - Reference to Subject Email using Subject Alternative Name

pedroigor:


matthewhelmke (dismissed):
Looks good to me.

vmuzikar:
@pedroigor Do you think we should explicitly mention the email's extracted the "rfc822Name" field of the SAN extension since there are other SAN fields as well?

pedroigor:
@vmuzikar, I'm not sure the name of the identity source is clearly stating that the email is extracted using the SAN extension ...

But if you think we should have it, I'm OK to include it ....

pedroigor:
@vmuzikar, done.

vmuzikar:
@pedroigor Thanks! I believe this makes it clear now.

matthewhelmke (approved):

pedroigor:
@matthewhelmke, code already pushed to repository. Can we have this change in ?

PR #335, opened 2018-03-07
KEYCLOAK-6800 added instructions for deleting users

matthewhelmke:
This new section parallels the phrasing and construction of the adding users section, so it should be pretty quick to review.


stianst:
Changes looks good to me, but there's some screw up with regards to commits. @matthewhelmke can you sort that out so there's only the single commit for adding delete user instructions?

stianst:
FIY https://github.com/keycloak/keycloak-documentation/pull/357 should fix the broken test

stianst (changes requested):
Fix commits

matthewhelmke:
Sorted

stianst (approved):

PR #336, opened 2018-03-07
KEYCLOAK-6084 attempting to help with Keycloak docs SEO

matthewhelmke:
This is a simple idea and strongly suggested by the resident RH experts as the best option for increasing search engine visibility for the Keycloak.org website and documentation. Worth a shot.

### DIFF
--- a/getting_started/topics/overview.adoc
+++ b/getting_started/topics/overview.adoc
@@ -7,5 +7,5 @@ and server configuration and does not cover any complex deployment options. If y
 in-depth discussion of any features or configuration options, consult the various reference guides available.
 
 ifeval::[{project_product}==true]
-{project_name} is based on the open source link:{https://www.keycloak.org/}[Keycloak] community project, which has its documentation link:{http://www.keycloak.org/documentation.html}[here]. 
+{project_name} is based on the open source link:https://www.keycloak.org/[Keycloak] community project, which has its documentation link:http://www.keycloak.org/documentation.html[here]. 
 endif::[]
--- a/getting_started/topics/overview.adoc
+++ b/getting_started/topics/overview.adoc
@@ -5,3 +5,7 @@ The purpose of this guide is to get you up and running as quickly as possible so
 play with and test-drive various features that {project_name} has. It relies heavily on the default database
 and server configuration and does not cover any complex deployment options. If you want a more
 in-depth discussion of any features or configuration options, consult the various reference guides available.
+
+ifeval::[{project_product}==true]
+{project_name} is based on the open source link:{https://www.keycloak.org/}[Keycloak] community project, which has its documentation link:{http://www.keycloak.org/documentation.html}[here]. 
+endif::[]


stianst:
I'm fairly certain this won't make any difference. Keycloak documentation results actually used to come before RH-SSO documentation results before we switched to single-page layout.

Did you look into how Google indexes large documents? Also, it seems that single-page layout even if the results show up is pretty much useless as the user will just be directed to the top of the first page rather than to the relevant page.

matthewhelmke:
I am not opposed to changing to using multi-page format in Keycloak documentation.

Were there any other changes to the Keycloak documentation as it was moved from the old to the new site? Or were the changes limited to moving and single-page layout? 

I'm asking because I think the combination of moving without the ability to set up redirects left the Google spiders with no way to keep the old Keycloak documentation ranking. Then, if the documentation changed from fairly unique to being the same as the RH docs at the same time, the Google algorithm may have decided it is just a copy site. This is why the link from the RH docs to the Keycloak docs could actually be helpful as it helps establish to the Google algorithm that the new Keycloak site is an authoritative source and not a potential spam/copied content site.

stianst:
There was two changes really. Multi page to single page and also moving from being hosted on Gitbook to keycloak.org.

If you take a look at http://www.keycloak.org/search.html that doesn't even return proper results for the docs and it's a custom search setup to search only our website, blog, mailing lists and documentation. So even though I think ranking and moving may have impacted the results on google.com there are also other issues.

stianst:
Searching a single big file is going to be useless though isn't it? Even if Google somehow finds the keys on the page it won't have a way to link to it.

matthewhelmke:
I'm getting a larger understanding here. So, this seems to encompass these three things:

1. Convincing Google to have its spiders index the new site. This is the only thing my current PR covers.
2. Changing the Keycloak.org site to use multi-page, so that Google has an easier time linking to specific content, because it isn't helpful to end users even if they receive a link from Google to the "right" page, but that page is so long they can't find what they were sent there to read.
3. Figuring out what is wrong with the internal search mechanism on Keycloak.org.

We could merge the current PR as a first step and move then to working on 2. Or, would you rather work on 1 and 2 and do a larger merge, hoping that helps with 3 while we reserve the right to do further work on that later?

stianst (approved):

stianst:
I believe 3 is caused fully by 2 and is basically just showing that Google can't index a large single page properly.

I'm fine with merging this for now 1. Sorting out 2 is going to be a pain as we need to use a different toolchain (asciidoc -> docbook -> html). This would mean figuring out again how we can add the menu to the top as well as how to add the edit/report links.

stianst (dismissed):
link:{https://www.keycloak.org/} should just be link:https://www.keycloak.org/. Same for docs link.

stianst:
Tests are failing due to syntax used in links.

stianst (approved):

PR #345, opened 2018-03-19
Add examples for action token spi to explain how to create a token.

FlxPeters:
I added two examples for a DemoActionToken implementation to the docs.

The most relevant part ist the hint for a private constructor without arguments. It's not mentioned in the docs yet and you get a confusing error message from the TokenValidator that it's not able to read the access token from JWT if you dont have this constructor.  

It took me several hours to find out. Maybe this example could help others to make their code work faster.


matthewhelmke (approved):
This looks good to me. Thank you.

hmlnarik (approved):
@FlxPeters Thanks for the contribution!

PR #349, opened 2018-03-27
[KEYCLOAK-4102] - Support lazy loading of paths via policy enforcer config

pedroigor:
Upstream: https://github.com/keycloak/keycloak/pull/5093

### DIFF
--- a/authorization_services/topics/enforcer-keycloak-enforcement-filter.adoc
+++ b/authorization_services/topics/enforcer-keycloak-enforcement-filter.adoc
@@ -96,7 +96,7 @@ Defines a URL where a client request is redirected when an "access denied" messa
 +
 ** *paths*
 +
-Specifies the paths to protect.
+Specifies the paths to protect. This configuration is optional. If not defined, the policy enforcer will discover all paths by fetching the resources you defined to your application in {project_name}, where these resources are defined with a `URI` representing some path in your application.
 +
 *** *name*
 +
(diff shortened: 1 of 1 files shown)


authorization_services/topics/enforcer-keycloak-enforcement-filter.adoc
```diff
@@ -96,7 +96,9 @@ Defines a URL where a client request is redirected when an "access denied" messa
 +
 ** *paths*
 +
-Specifies the paths to protect.
+Specifies the paths to protect. This configuration is optional and if not defined the policy enforcer is going to
+discover all paths that should be protected by fetching the resources you defined to your application in {project_name}, where these resources are defined
+with a `URI` representing some path in your application.
 +
```
matthewhelmke: Question about "should be protected": how is it known what "should" be protected? Is it more accurate to say "may be protected" or, in that case, perhaps leave the phrase out entirely?

Possible edit, maybe with [ ] phrase deleted:
Specifies the paths to protect. This configuration is optional. If not defined, the policy enforcer will discover all paths [that should be protected] by fetching the resources you defined to your application in {project_name}, where these resources are defined with a `URI` representing some path in your application.

matthewhelmke (changes requested):

matthewhelmke (approved):

pedroigor:
@matthewhelmke, upstream merged. Tks.

PR #355, opened 2018-04-02
Fix copied&pasted heading

Yonezawa-T2:
There are two `resource_registration_endpoint` headings, but one should be `permission_endpoint`.


matthewhelmke (approved):
Thank you!

pedroigor (approved):

PR #360, opened 2018-04-17
[RHSSO-1242] and [RHSSO-1243]

matthewhelmke:
Files display properly in local builds and I do not see what is mentioned in the reopened issues. Still, I noticed and fixed inconsistent spacing at top of three social IdP files and fixed that before trying, so perhaps that was the issue.


hmlnarik (approved):
Let's try this fix.

PR #365, opened 2018-04-19
[BZ1567846] Added location of RPM Based installation start script

jparrill:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1567846. As you see on https://access.redhat.com/documentation/en-us/red_hat_single_sign-on/7.1/html/server_installation_and_configuration_guide/operating-mode there is no clue about the binaries location, this MR solves that on the RPM installation based.

### DIFF
--- a/getting_started/topics/first-boot/boot.adoc
+++ b/getting_started/topics/first-boot/boot.adoc
@@ -8,7 +8,13 @@ image:{project_images}/standalone-boot-files.png[]
 
 To boot the server:
 
-.Linux/Unix
+.Linux/Unix (rpm)
+[source]
+----
+$ /opt/rh/eap<EAP_VERSION>/root/usr/share/<VERSION_NAME>/bin/standalone.sh
+----
+
+.Linux/Unix (zip|tar)
 [source]
 ----
 $ .../bin/standalone.sh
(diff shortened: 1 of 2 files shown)


matthewhelmke (approved):
Thank you!

getting_started/topics/first-boot/boot.adoc
```diff
@@ -8,7 +8,13 @@ image:{project_images}/standalone-boot-files.png[]
 
 To boot the server:
 
-.Linux/Unix
+.Linux/Unix (rpm)
```
stianst, line 11: These docs are shared between Keycloak and RH-SSO. RPMs are only available for RH-SSO, but this change will show steps for RPMs in both sets of docs. It would be better to put a new paragraph above (after go to the bin directory of the server distribution) and wrap that in a if condition to only display for RH-SSO.

server_installation/topics/operating-mode/standalone.adoc
```diff
@@ -17,7 +17,13 @@ image:{project_images}/standalone-boot-files.png[]
 
 To boot the server:
 
-.Linux/Unix
+.Linux/Unix (rpm)
```
stianst, line 20: Same comment as above

stianst (changes requested):

stianst:
Closing this PR as it's not correct and didn't get any feedback from contributor.

PR #367, opened 2018-04-20
Fix typo in server_admin

RemiEven:


matthewhelmke (approved):
Thank you! From a language standpoint, the change is simple and easy to approve. @stianst, can you please have someone review from an engineering perspective?

server_admin/topics/admin-cli.adoc
```diff
@@ -1084,7 +1084,7 @@ Use a dedicated [command]`add-roles` command to add realm roles to a user.
 Use the following example to add the [command]`user` role to user [command]`testuser`.
 [options="nowrap"]
 ----
-$ kcadm.sh add-roles --username testuser --rolename user -r demorealm
+$ kcadm.sh add-roles --uusername testuser --rolename user -r demorealm
```
stianst, line 1087: This doesn't look right. Surely the param name isn't "--uusername", but "--username" as it is in the docs currently

RemiEven:
When I tried the commands without the double 'u', they failed. Moreover the help from the command line (kcadm.sh help add-roles) explicitly mentions "uusername" multiple times.

stianst:
In that case I'd say the typo is in the CLI and not the docs. It makes no sense that the param is uusername

PR #376, opened 2018-05-03
Error about how to run host-master in Windows  enviroment

carlos-perez-rp:
small error, instead of running master, run slave has been described in Windows section.


matthewhelmke (approved):
Nice catch! Thank you for the correction.

stianst (approved):

PR #380, opened 2018-05-09
KEYCLOAK-7241 providers directory no longer exists

slominskir:
The providers directory has been removed so use the standalone/deployments directory instead (KEYCLOAK-7241).


matthewhelmke (approved):
This looks good to me, thank you!

stianst (approved):

PR #382, opened 2018-05-14
[CLOUD-2437] Describe JGroups ASYM_ENCRYPT default setting for x509 templates

iankko:
Mention JGroups ASYM_ENCRYPT cluster traffic encryption protocol, which is the default setting from now on for the templates using the re-encrypt TLS termination

Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>


iankko:
@matthewhelmke @stianst PTAL (once you got a chance)

Thank you, Jan

matthewhelmke (approved):
This looks good to me. Thank you!

PR #383, opened 2018-05-15
KEYCLOAK-6771 Holder of Key mechanism: OAuth 2.0 Mutual TLS Client Authentication and Certificate Bound Access Tokens

tnorimat:


tnorimat:
This documentation is for Mutual TLS Certificate Bound Access Token (https://github.com/keycloak/keycloak/pull/5083).

tnorimat:
Some errors occurred, but it seems that this documentation has nothing to do with them due to their happening in Authorization Service documentation check.

server_admin/topics/clients/client-oidc.adoc
```diff
@@ -123,10 +123,30 @@ See link:{adapterguide_link}[{adapterguide_name}] for more information.
...
+
+To enable mutual TLS in {project_name}, please see <<_enable-mtls-wildfly, Enable mutual SSL in WildFly>>.
+
+In the following cases, {project_name} verify whether the client sending the access token or the refresh token is the genuine holder of them. If not, {project_name} rejects them.
+
+* When a token refresh request is sent onto Token Endpoint with a Holder-of-Key type refresh token.
+* When a UserInfo request is sent onto UserInfo Endpoint with a Holder-of-Key type access token.
+* When a logout request is sent onto Logout Endpoint with a Holder-of-Key type refresh token.
+
+Please see https://tools.ietf.org/html/draft-ietf-oauth-mtls-08#section-3[Mutual TLS Client Certificate Bound Access Tokens] in the OAuth 2.0 Mutual TLS Client Authentication and Certificate Bound Access Tokens for more details.
```
matthewhelmke: ==== Advanced Settings

[[_mtls-client-certificate-bound-tokens]]
*OAuth 2.0 Mutual TLS Client Certificate Bound Access Token*

Mutual TLS binds an access token and a refresh token with a client certificate exchanged during TLS handshake. This prevents an attacker who finds a way to steal these tokens from exercising the tokens. The type of token is called a holder-of-key token. Unlike bearer tokens, the recipient of a holder-of-key token can verify whether the sender of the token is legitimate.

If the following conditions are satisfied on a token request, {project_name} will bind an access token and a refresh token with a client certificate and issue them as holder-of-key tokens. If all conditions are not met, {project_name} rejects the token request.

* The feature is turned on
* A token request is sent to the token endpoint in an authorization code flow or a hybrid flow
* On TLS handshake, {project_name} requests a client certificate and a client send its client certificate
* On TLS handshake, {project_name} successfully verifies the client certificate

To enable mutual TLS in {project_name}, see <<_enable-mtls-wildfly, Enable mutual SSL in WildFly>>.

In the following cases, {project_name} will verify the client sending the access token or the refresh token; if verification fails, {project_name} rejects the token.

* A token refresh request is sent to the token endpoint with a holder-of-key refresh token
* A UserInfo request is sent to UserInfo endpoint with a holder-of-key access token
* A logout request is sent to Logout endpoint with a holder-of-key refresh token

Please see https://tools.ietf.org/html/draft-ietf-oauth-mtls-08#section-3[Mutual TLS Client Certificate Bound Access Tokens] in the OAuth 2.0 Mutual TLS Client Authentication and Certificate Bound Access Tokens for more details.

server_admin/topics/threat/compromised-tokens.adoc
```diff
@@ -6,6 +6,8 @@ There are a few things you can do to mitigate access tokens and refresh tokens f
 Another thing you can do to mitigate leaked access tokens is to shorten their lifespans.  You can specify this within the <<_timeouts, timeouts page>>.
 Short lifespans (minutes) for access tokens for clients and applications to refresh their access tokens after a short amount of time. If an admin detects a leak, they can logout all user sessions to invalidate these refresh tokens or set up a revocation policy. Making sure refresh tokens always stay private to the client and are never transmitted ever is very important as well.
 
+Other way of mitigating leaked access tokens and refresh tokens is to issue these tokens as Holder-of-Key token. <<_mtls-client-certificate-bound-tokens, OAuth 2.0 Mutual TLS Client Certificate Bound Access Token>> shows how to issue these tokens as Holder-of-Key token instead of bearer token.
+
```
matthewhelmke: You can also mitigate against leaked access tokens and refresh tokens by issuing these tokens as holder-of-key tokens. See <<_mtls-client-certificate-bound-tokens, OAuth 2.0 Mutual TLS Client Certificate Bound Access Token>> to learn how.

matthewhelmke (changes requested):
I have cleaned up the language in my comments.

tnorimat:
Thank you for your comments. I've incorporated them.

matthewhelmke (dismissed):

matthewhelmke:
Thank you!

server_admin/topics/clients/client-oidc.adoc
```diff
@@ -123,10 +123,29 @@ See link:{adapterguide_link}[{adapterguide_name}] for more information.
 To fill in the `Web Origins` data, enter in a base URL and click the + sign to add.  Click the - sign next to URLs you want to remove.
 Remember that you still have to click the `Save` button!
 
+==== Advanced Settings
 
+[[_mtls-client-certificate-bound-tokens]]
+*OAuth 2.0 Mutual TLS Client Certificate Bound Access Token*
 
+Mutual TLS binds an access token and a refresh token with a client certificate exchanged during TLS handshake. This prevents an attacker who finds a way to steal these tokens from exercising the tokens. The type of token is called a holder-of-key token. Unlike bearer tokens, the recipient of a holder-of-key token can verify whether the sender of the token is legitimate.
```
mposolda, line 132: Is it possible to add WARNING here that our client adapters currently doesn't do this verification, but we plan to add it in the future? 

I want to discuss with you first, but I believe it will be good if our client adapters have support for this? I think the specs mentions it on few places that ResourceServer should verify the access tokens and IMO this will be one of important advantages of this feature. WDYT? We can create separate JIRA for this and it can be done as followup once this PR is merged.

server_admin/topics/clients/client-oidc.adoc
```diff
@@ -123,10 +123,29 @@ See link:{adapterguide_link}[{adapterguide_name}] for more information.
...
+* On TLS handshake, {project_name} successfully verifies the client certificate
+
+To enable mutual TLS in {project_name}, see <<_enable-mtls-wildfly, Enable mutual SSL in WildFly>>.
+
+In the following cases, {project_name} will verify the client sending the access token or the refresh token; if verification fails, {project_name} rejects the token.
+
+* A token refresh request is sent to the token endpoint with a holder-of-key refresh token
+* A UserInfo request is sent to UserInfo endpoint with a holder-of-key access token
+* A logout request is sent to Logout endpoint with a holder-of-key refresh token
+
```
mposolda: Is it possible to add token Introspection endpoint (I've mentioned that in the codebase PR) ?

mposolda (changes requested):
@tnorimat Thanks a lot! I've added 2 inline comments for the docs.

tnorimat:
@mposolda 
Thanks a lot for your comments.
As for client adapter, I would like to consider it and make comments afterwards.
As for Token Introspection Endpoint,  I would like to follow your comment.

mposolda:
@tnorimat I've just deleted the inline comment about token introspection as we agreed on the "implementation" PR. Cool! Sure, we can discuss client adapter later. Thanks a lot for the contributions!

tnorimat:
About client adapter's support for HoK verification, Yes, I agree with you. In this PR, I will add WARNING to mention that no keycloak's adapter currently verifies HoK Token.
And in the future, I would like to try to contribute HoK Token verification on keycloak's wildfly/Jboss EAP client adapter in operating as Resource Server (namely, "bearer-only" : true in keycloak.json).
Therefore, at that time, I would like to issue another JIRA ticket for treating it.
Is it OK? If so, I would like to propose the following description. 

WARNING:
Currently not any keycloak's client adapter supports this holder-of-key token verification.
Therefore, every keycloak's client adapter treats access and refresh tokens as bearer tokens.

mposolda:
@tnorimat +1 for add the warning like this.

@matthewhelmke Is the WARNING from the previous comment ok?

@tnorimat That would be great if you create JIRA and link it with the KEYCLOAK-6771 as a follow-up JIRA. 

Since we're discussing already, I have few points:
- Maybe we can have a boolean flag to add to adapter configuration (keycloak.json) whether the access token should be verified just as a bearer token OR if the holder-of-key verification is required? I was thinking whether to combine both (For example you can have set of more "trusted"clients, where is OK to have just bearer verification, but for some other clients require hok verification), but I guess that in first stage, it's ok to have just boolean flag and eventually add more fine-grain configuration later if someone needs it. WDYT?
- Te flag shouldn't be likely specific just to bearer-only clients. That's because even normal clients can be authenticated through bearer token (BearerTokenRequestAuthenticator), so even for those clients it makes sense to support this flag.

tnorimat:
I've created a new JIRA ticket as you said. Please refer to the following link.
https://issues.jboss.org/browse/KEYCLOAK-7499

And I would like to consider points you just mentioned above.

matthewhelmke:
@tnorimat @mposolda 
This is how I would write the warning.

WARNING:
None of the keycloak client adapters currently support holder-of-key token verification.
Instead, keycloak adapters currently treat access and refresh tokens as bearer tokens.

tnorimat:
I've add the sentence @matthewhelmke corrected and squashed all commits. Could you please confirm this commit?

mposolda (approved):
@tnorimat Thanks!

mposolda:
@matthewhelmke FYI. I've created separate JIRA for the issue, which is now causing build to fail: https://issues.jboss.org/browse/KEYCLOAK-7506

PR #408, opened 2018-06-18
Update cache.adoc

zyv:
Hopefully the cache should _de_crease response times, otherwise, I guess it's better to do without it :)


matthewhelmke (approved):
Thank you.

PR #410, opened 2018-06-20
KEYCLOAK-6781 Docs for upgrading server to Wildfly 13 and upgrading J…

mposolda:
…DG to 7.2

### DIFF
--- a/server_installation/topics/cache/eviction.adoc
+++ b/server_installation/topics/cache/eviction.adoc
@@ -5,61 +5,31 @@ There are multiple different caches configured for {project_name}.
 There is a realm cache that holds information about secured applications, general security data, and configuration options.
 There is also a user cache that contains user metadata.  Both caches default to a maximum of 10000 entries and use a least recently used eviction strategy.
 Each of them is also tied to an object revisions cache that controls eviction in a clustered setup.
-This cache is created implicitly and has twice the configured size.
-There are also separate caches for user sessions, offline tokens, and login failures.  These caches are unbounded in size as well.
+This cache is created implicitly and has twice the configured size. There are some more caches for save the authorization data, keys etc.
+There are also separate caches for user sessions, offline tokens, login failures and other things. These caches are unbounded in size by default.
 
 The eviction policy and max entries for these caches can be configured in the _standalone.xml_, _standalone-ha.xml_, or
-_domain.xml_ depending on your <<_operating-mode, operating mode>>.
+_domain.xml_ depending on your <<_operating-mode, operating mode>>. In the configuration file, there is the part with infinispan
+subsystem, which looks similar to this:
 
-.non-clustered
-[source,xml]
+[source,xml,subs="attributes+"]
 ----
-<subsystem xmlns="urn:jboss:domain:infinispan:4.0">
-    <cache-container name="keycloak" jndi-name="infinispan/Keycloak">
+<subsystem xmlns="{subsystem_infinispan_xml_urn}">
+    <cache-container name="keycloak">
         <local-cache name="realms">
-            <eviction max-entries="10000" strategy="LRU"/>
+            <object-memory size="10000"/>
         </local-cache>
         <local-cache name="users">
-            <eviction max-entries="10000" strategy="LRU"/>
+            <object-memory size="10000"/>
         </local-cache>
-        <local-cache name="sessions"/>
-        <local-cache name="offlineSessions"/>
-        <local-cache name="loginFailures"/>
-        <local-cache name="work"/>
-        <local-cache name="authorization">
-           <eviction strategy="LRU" max-entries="100"/>
-        </local-cache>
+        ...
         <local-cache name="keys">
-            <eviction strategy="LRU" max-entries="1000"/>
+            <object-memory size="1000"/>
             <expiration max-idle="3600000"/>
         </local-cache>
+        ...
     </cache-container>
 ----
 
-
-.clustered
-[source,xml]
-----
-<subsystem xmlns="urn:jboss:domain:infinispan:4.0">
-    <cache-container name="keycloak" jndi-name="infinispan/Keycloak">
-        <transport lock-timeout="60000"/>
-        <local-cache name="realms">
-            <eviction max-entries="10000" strategy="LRU"/>
-        </local-cache>
-        <local-cache name="users">
-            <eviction max-entries="10000" strategy="LRU"/>
-        </local-cache>
-        <distributed-cache name="sessions" mode="SYNC" owners="1"/>
-        <distributed-cache name="offlineSessions" mode="SYNC" owners="1"/>
-        <distributed-cache name="loginFailures" mode="SYNC" owners="1"/>
-        <distributed-cache name="authorization" mode="SYNC" owners="1"/>
-        <replicated-cache name="work" mode="SYNC"/>
-        <local-cache name="keys">
-            <eviction max-entries="1000" strategy="LRU"/>
-            <expiration max-idle="3600000"/>
-        </local-cache>
-    </cache-container>
-----
-
-To limit or expand the number of allowed entries simply add or edit the `eviction` element or the `expiration` element of particular cache
+To limit or expand the number of allowed entries simply add or edit the `object` element or the `expiration` element of particular cache
 configuration. 
--- a/server_installation/topics/cache/eviction.adoc
+++ b/server_installation/topics/cache/eviction.adoc
@@ -1,12 +1,15 @@
 
+[[_eviction]]
 === Eviction and Expiration
 
 There are multiple different caches configured for {project_name}.
 There is a realm cache that holds information about secured applications, general security data, and configuration options.
 There is also a user cache that contains user metadata.  Both caches default to a maximum of 10000 entries and use a least recently used eviction strategy.
 Each of them is also tied to an object revisions cache that controls eviction in a clustered setup.
-This cache is created implicitly and has twice the configured size. There are some more caches for save the authorization data, keys etc.
-There are also separate caches for user sessions, offline tokens, login failures and other things. These caches are unbounded in size by default.
+This cache is created implicitly and has twice the configured size. The same applies for the `authorization` cache, which holds
+the authorization data. The `keys` cache holds data about external keys and does not need to have dedicated revisions cache. Rather
+it has `expiration` explicitly declared on it, so the keys are periodically expired and forced to be periodically downloaded from external clients or
+identity providers.
 
 The eviction policy and max entries for these caches can be configured in the _standalone.xml_, _standalone-ha.xml_, or
 _domain.xml_ depending on your <<_operating-mode, operating mode>>. In the configuration file, there is the part with infinispan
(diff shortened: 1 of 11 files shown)


server_installation/topics/cache/eviction.adoc
```diff
@@ -5,61 +5,31 @@ There are multiple different caches configured for {project_name}.
 There is a realm cache that holds information about secured applications, general security data, and configuration options.
 There is also a user cache that contains user metadata.  Both caches default to a maximum of 10000 entries and use a least recently used eviction strategy.
 Each of them is also tied to an object revisions cache that controls eviction in a clustered setup.
-This cache is created implicitly and has twice the configured size.
-There are also separate caches for user sessions, offline tokens, and login failures.  These caches are unbounded in size as well.
+This cache is created implicitly and has twice the configured size. There are some more caches for save the authorization data, keys etc.
+There are also separate caches for user sessions, offline tokens, login failures and other things. These caches are unbounded in size by default.
```
matthewhelmke: The rewrite below is better. Even better yet would be if we could fill in details instead of saying "and so on" or "and other things."

This cache is created implicitly and has twice the configured size. There are also caches for the authorization data, keys, and so on. In addition, there are also separate caches for user sessions, offline tokens, login failures, and other things. These caches are unbounded in size by default.

server_installation/topics/cache/replication.adoc
```diff
@@ -1,7 +1,8 @@
 
 === Replication and Failover
 
-The `sessions`, `authenticationSessions`, `offlineSessions` and `loginFailures` caches are the only caches that may perform replication.  Entries are
+There are some caches like `sessions`, `authenticationSessions`, `offlineSessions`, `loginFailures` and few others, which are configured
+as distributed caches in case of clustered setup.  Entries are
```
matthewhelmke: Again, if details are available, it would be nice to replace "and a few others" with those details. If no, this rewrite will be sufficient.

There are caches like `sessions`, `authenticationSessions`, `offlineSessions`, `loginFailures` and a few others, which are configured as distributed caches when using a clustered setup. Entries are...

server_installation/topics/cache/replication.adoc
```diff
@@ -23,3 +24,7 @@ Here we've changed it so at least two nodes will replicate one specific user log
 TIP:  The number of owners recommended is really dependent on your deployment.  If you do not care if users are logged
       out when a node goes down, then one owner is good enough and you will avoid replication.
 
+TIP:  It is very good if you configure your environment to use loadbalancer with sticky sessions. It is beneficial for performance
+      as {project_name} server, where the particular request is served, will be usually the owner of the data from distributed cache and
+      hence it will be able to lookup the data locally. More details in <<sticky-sessions>>.
+
```
matthewhelmke: TIP:  It is generally wise to configure your environment to use loadbalancer with sticky sessions. It is beneficial for performance as {project_name} server, where the particular request is served, will be usually the owner of the data from the distributed cache and will therefore be able to look up the data locally. See <<sticky-sessions>> for more details.

server_installation/topics/clustering/troubleshooting.adoc
```diff
@@ -1,19 +1,22 @@
 
 === Troubleshooting
 
-Note that when you run cluster, you should see message similar to this in the log of both cluster nodes:
-
+* Note that when you run cluster, you should see message similar to this in the log of both cluster nodes:
++
```
matthewhelmke: ...run a cluster...

server_installation/topics/clustering/troubleshooting.adoc
```diff
@@ -1,19 +1,22 @@
...
 Usually it's best practice to have your cluster nodes on private network without firewall for communication among them.
 Firewall could be enabled just on public access point to your network instead.
 If for some reason you still need to have firewall enabled on cluster nodes, you will need to open some ports.
 Default values are UDP port 55200 and multicast port 45688 with multicast address 230.0.0.4.
 Note that you may need more ports opened if you want to enable additional features like diagnostics for your JGroups stack.
 {project_name} delegates most of the clustering work to Infinispan/JGroups.
 For more information, see link:{appserver_jgroups_link}[JGroups] in the _{appserver_jgroups_name}_.
+
+* If you are interested about failover support (hight availability), evictions, expiration and cache tunings, you can go to the
+<<cache-configuration>>.
```
matthewhelmke: * If you are interested in failover support (high availability), evictions, expiration and cache tuning, see ...

server_installation/topics/operating-mode/crossdc.adoc
```diff
@@ -233,7 +235,12 @@ Details of this more-detailed setup are out-of-scope of the {project_name} docum
 NOTE: Details about the configuration options inside `replicated-cache-configuration` are explained in <<tuningcache>>, which includes information about tweaking some of those options.
 
 +  
-. Newer {jdgserver_name} releases require authorization before accessing protected caches over network. {project_name} requires updates to
+. Some {jdgserver_name} server releases require authorization before accessing protected caches over network.
++
+NOTE: You shouldn't see any issue if you use recommended {jdgserver_name} {jdgserver_version} server and this step can (and should) be ignored.
```
matthewhelmke: s/shouldn't/should not

upgrading/topics/keycloak/changes.adoc
```diff
@@ -1,5 +1,25 @@
...
+Configuration changes::
+  There are few configuration changes in the `standalone(-ha).xml` and `domain.xml` files. You should follow section <<_install_new_version>>
+  to handle the migration of configuration files automatically. Just in case you want more details (For example if you did some configuration
+  changes on your own), here is the list of the most important changes in the configuration:
+  * Element `eviction` on infinispan caches is now deprecated (unused) and was replaced by element `object-memory`
+  * There is no `jndi-attribute` on `cache-container` element in infinispan subsystem.
+  * If you use cross-dc integration, there is a need to add property `protocolVersion` with value `2.6` on `remote-store` . This is because
+  hotrod client ({project_name} server) is using infinispan 9 based on newer version of hotrod protocol than {jdgserver_name} server, which
+  is still using infinispan 8.
+
```
matthewhelmke: === Migrating to 4.1.0

==== Upgrade to Wildfly 13

The {project_name} server was upgraded to use Wildfly 13 as the underlying container. This does not directly involve any specific {project_name} server functionality, but there are few changes related to the migration, which worth mentioning.

Dependency updates::
  The dependencies were updated to the versions used by Wildfly 13 server. For example, Infinispan is now 9.2.4.Final and Undertow is 2.0.9.Final

Configuration changes::
  There are few configuration changes in the `standalone(-ha).xml` and `domain.xml` files. You should follow the <<_install_new_version>> section
  to handle the migration of configuration files automatically. If more detail is needed, because, for example, you did some configuration
  changes on your own, here is a list of the most important changes:
  * Element `eviction` on infinispan caches is now deprecated (unused) and was replaced by element `object-memory`
  * There is no `jndi-attribute` on `cache-container` element in infinispan subsystem.
  * If you use cross-dc integration, you must add property `protocolVersion` with value `2.6` on `remote-store` . This is because
  the hotrod client ({project_name} server) is using infinispan 9 based on newer version of hotrod protocol than {jdgserver_name} server, which
  is still using infinispan 8.

matthewhelmke (changes requested):
Looks pretty good, just requesting some clean up.

server_installation/topics/cache/replication.adoc
mposolda: @matthewhelmke Well, I am not sure. Explicitly mention all the caches is more clear, but it has also one disadvantage.  Always when we add new cache into the configuration, we will also need to update this documentation with all the references to cache names. If you write "and few others", you explicitly mention to admin that there are more caches and he will quickly figure out by looking  to the configuration file what the caches are... :) WDYT? For now, I've just used the sentence you mentioned with "a few others" in it.

server_installation/topics/cache/eviction.adoc
mposolda: I've changed like this in my local fork and added some details. Although adding more details is always double-side coin as it means more documentation changes will be needed in the future when implementation changes. See my comment below. So I've used this:

This cache is created implicitly and has twice the configured size. The same applies for the `authorization` cache, which holds for
the authorization data. The `keys` cache holds data about external keys and doesn't need to have dedicated revisions cache. Rather
it has `expiration` explicitly declared on it, so the keys are forced to be periodically downloaded from external clients or
identity providers.

In addition, there are also separate caches for user sessions, client sessions, offline tokens, login failures, action tokens and other things.
These caches are unbounded in size by default.

matthewhelmke: Perhaps we can create a list of all caches and put that list in an obvious place in the documentation, then refer to it as the canonical reference when we make comments like "and other things".

My fear is that we don't want to frustrate readers/users by being vague. I definitely also agree with you about the double-sided coin of more detail means more documentation changes. The balance is that we should provide the information our users need, and only that information when we are writing procedures, but we should also find a way to provide a complete reference somewhere to the other details that they may eventually need.

server_installation/topics/cache/replication.adoc
matthewhelmke: See my comment above. It is not always an easy balance, and you probably know better than I what our users must know versus what they would like to have as reference.

server_installation/topics/cache/eviction.adoc
mposolda: Yes, I see. I agree with you that it is sometimes not easy to find the right balance.. In the end, I've added even more details into the "evictions" sections related to caches (talked today with @pedroigor about some things, which we're missing the docs). Also all the caches are used here and referenced from "replication" chapter.

mposolda:
@matthewhelmke Thanks again for the review! I've updated the documentation per your feedback. But I also needed to do some updates to the content itself due the updates in the Wildfly 13 integration, which I needed to change in the meantime in the implementation PR: https://github.com/keycloak/keycloak/pull/5293 . I also added some more details to the "eviction" chapter as we discussed.

So new review will be probably needed. 

I've did all the changes into the separate second commit. Maybe it's easier for you to review it better to see all the new changes.

server_installation/topics/cache/eviction.adoc
matthewhelmke: I like that approach, @mposolda. What do you think about replacing "and other things" with a reference like "and others, for more information about all of the caches, see ..."?

matthewhelmke:
@mposolda, thank you. I will do another review later today.

server_installation/topics/cache/eviction.adoc
```diff
@@ -32,4 +35,24 @@ subsystem, which looks similar to this:
...
-configuration. 
\ No newline at end of file
+configuration.
+
+In addition, there are also separate caches `sessions`, `clientSessions`, `offlineSessions`, `offlineClientSessions`,
+`loginFailures` and `actionTokens`. These caches are distributed in cluster environment and they are unbounded in size by default.
+If they are bounded, it would then be possible that some sessions will be lost. Expired sessions are cleared internally
+by {project_name} itself to avoid growing the size of these caches
+without limit. If you see memory issues due the big amount of sessions, you can try to:
+
```
matthewhelmke: If you see memory issues due to a large number of sessions...

server_installation/topics/cache/eviction.adoc
```diff
@@ -32,4 +35,24 @@ subsystem, which looks similar to this:
...
+* Increase the size of cluster (more nodes in cluster means that sessions are spread more equally among nodes)
+
+* Increase the memory for {project_name} server process
+
+* Lower the amount of owners to ensure that caches are saved just on single place. See <<_replication>> for more details
```
matthewhelmke: * Decrease the number of owners to ensure that caches are saved in one single place.

server_installation/topics/cache/eviction.adoc
```diff
@@ -32,4 +35,24 @@ subsystem, which looks similar to this:
...
+
+* Disable l1-lifespan for distributed caches. See Infinispan documentation for more details
+
+* Decrease session timeouts, which could be done individually for each realm in {project_name} admin console. But this could affect
+usability for end users. See link:{adminguide_timeouts_link}[{adminguide_timeouts_name}] for more details.
+
+There is replicated cache `work`, which is mostly used to send messages among cluster nodes and it is also unbounded by default. However
+this cache should not cause any memory issues as entries in this cache are very short lived.
```
matthewhelmke: There is an additional replicated cache, `work`, which is mostly used to send messages among cluster nodes; it is also unbounded by default. However, this cache should not cause any memory issues as entries in this cache are very short-lived.

matthewhelmke (changes requested):
I have some very minor edits in your changes. Overall, I like these changes very much!

mposolda:
@matthewhelmke I've finally updated per your feedback. Sorry it took so long... :(

matthewhelmke (approved):
Thank you!

PR #416, opened 2018-06-27
adding template information to contributors guide

matthewhelmke:
The templates should be called out in the contributor's guide more clearly as we are trying to clean up our overall writing standards and style.

### DIFF
--- a/internal_resources/contributing.adoc
+++ b/internal_resources/contributing.adoc
@@ -3,7 +3,7 @@
 
 This guide will help new contributors use `git` and GitHub for Keycloak documentation submissions and suggestions.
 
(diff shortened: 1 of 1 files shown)


internal_resources/contributing.adoc
```diff
@@ -3,6 +3,8 @@
 
 This guide will help new contributors use `git` and GitHub for Keycloak documentation submissions and suggestions.
 
+Before you start, please take a look at our writing templates contained in the link:https://github.com/keycloak/keycloak-documentation/tree/master/internal_resources[internal resources directory], which also describes each template file with information about when to use which. We are working to update our documentation to match the style of these templates and request that new contributions please use the format and style in them.
+
```
tanberry: Matthew, let's tweak  that last sentence say something like:

"We are working to modularize our documentation to put procedural content, conceptual content, and reference content into separate .adoc files, and we kindly request that new contributions please use these templates." 
I know it is the same meaning, but it is worded more strongly. Thanks for adding the link!

matthewhelmke: That is a reasonable suggestion, @tanberry. Thank you.

stianst (approved):

PR #418, opened 2018-06-29
[KEYCLOAK-7062] - Group Claims should be optional

pedroigor:
Upstream https://github.com/keycloak/keycloak/pull/5334


matthewhelmke (approved):
This looks good to me. Thanks, @pedroigor

PR #434, opened 2018-07-24
Update release notes for 4.2.0.Final

mposolda:


release_notes/topics/4_2_0_final.adoc
```diff
@@ -3,4 +3,15 @@
...
+
+= SAML adapter multitenancy support
+
+The SAML adapter can support multi-tenancy now just like the built in adapter for OpenID Connect.
+
+
+= An option to create claims with dots (.) in them
+
+In previous versions, it was not possible to create claims in the token with the claim name containing dot (.) character. Now it is
+possible to escape the dot character in the configuration, so claim name with the dot character can be used.
```
matthewhelmke: One minor suggestion:

In previous versions, it was not possible to create claims in the token using a claim name containing a dot (.) character. Now it is possible to escape the dot character in the configuration, so a claim name with the dot character can be used.

matthewhelmke (changes requested):

mposolda:
@matthewhelmke Thanks for the review! I've updated per your feedback.

matthewhelmke (approved):

PR #435, opened 2018-07-24
KEYCLOAK-7270

slominskir:
Support for automatically linking brokered identities
https://issues.jboss.org/browse/KEYCLOAK-7270

### DIFF
--- a/server_admin/topics/identity-broker/first-login-flow.adoc
+++ b/server_admin/topics/identity-broker/first-login-flow.adoc
@@ -64,3 +64,16 @@ Verify Existing Account By Re-authentication::
   User can also re-authenticate with some different identity provider, which is already linked to his {project_name} account.
   You can also force users to use OTP. Otherwise it's optional and used only if OTP is already set for the user account.
 
+==== Automatically Link Existing First Login Flow
+WARNING: The AutoLink authenticator would be dangerous in a generic environment where users can register themselves using arbitrary usernames/email addresses. Do not use this authenticator unless registration of users is carefully curated and usernames/email addresses are assigned, not requested.
+
+In order to configure a first login flow in which users are automatically linked without being prompted, create a new flow with the following two authenticators:
+
+Create User If Unique::
+This authenticator ensures unique users are handled. Set the authenticator requirement to "Alternative".
+
+Automatically Link Existing Account::
+Automatically link brokered identities without any validation with this authenticator. This is useful in an intranet environment of multiple user databases each with overlapping usernames/email addresses, but different passwords, and you want to allow users to use any password without having to validate. This is only reasonable if you manage all internal databases, and usernames/email addresses from one database matching those in another database belong to the same person. Set the authenticator requirement to "Alternative".
+
+NOTE: The described setup uses two authenticators, and is the simplest one, but it is possible to use other
+authenticators according to your needs. For example, you can add the Review Profile authenticator to the beginning of the flow if you still want end users to confirm their profile information.


server_admin/topics/identity-broker/first-login-flow.adoc
```diff
@@ -64,3 +64,11 @@ Verify Existing Account By Re-authentication::
...
+In order to configure an alternative first login flow in which users are automatically linked without prompt create a new flow with the following two authenticators:
+
+Create User If Unique::
+  This authenticator ensures unique users are handled.  Set the authenticator requirement to "Alternative".
+
+Automatically Link Existing Account::
+  Automatically link brokered identities without any validation with this authenticator. This is useful in an intranet environment of multiple user databases each with overlapping usernames/email addresses, but different passwords and you want to allow users to use any password without having to validate.  This is only reasonable if you manage all internal databases, and usernames/email addresses from one database matching those in another database belong to the same person.  This authenticator would be dangerous in a generic environment where users can register themselves using arbitrary usernames / email addresses.  One more time - don't use this unless registration of users is carefully curated and usernames/email addresses are assigned, not requested.  Set the authenticator requirement to "Alternative".
```
mposolda: Is it possible to convert the whole text starting with sentence "This authenticator would be dangerous in generic ..." to WARNING box? I think it can be done just by starting it with next paragraph prefixed with the "WARNING: " . Something like:

```
WARNING: This authenticator would be dangerous in generic ...
```

server_admin/topics/identity-broker/first-login-flow.adoc
```diff
@@ -64,3 +64,11 @@ Verify Existing Account By Re-authentication::
   User can also re-authenticate with some different identity provider, which is already linked to his {project_name} account.
   You can also force users to use OTP. Otherwise it's optional and used only if OTP is already set for the user account.
 
+==== Automatically Link Existing First Login Flow
+In order to configure an alternative first login flow in which users are automatically linked without prompt create a new flow with the following two authenticators:
```
mposolda: Is it possible to add small note that it's still possible to use the chain with other authenticators? For example something like:

```
NOTE: The described setup with two authenticators is the simplest one, but it is possible to chain with other 
authenticators according to your needs. For example you can add `Review Profile` authenticator to the beginning if 
you still want end users to confirm their profile informations.
```

What do you think?

mposolda (changes requested):
Thanks a lot for the PR! Few minor comments below:

slominskir:
Your suggestions sound good.  I've updated the request by adding another commit and then squashing it into one.

mposolda (dismissed):
@slominskir Thanks! Looks nice to me. IMO this is ready after also the implementation PR will be addressed (test added).

I am adding @matthewhelmke (our documentation expert) for more documentation review.

server_admin/topics/identity-broker/first-login-flow.adoc
```diff
@@ -64,3 +64,16 @@ Verify Existing Account By Re-authentication::
...
+In order to configure a first login flow in which users are automatically linked without prompt create a new flow with the following two authenticators:
+
+Create User If Unique::
+  This authenticator ensures unique users are handled.  Set the authenticator requirement to "Alternative".
+
+Automatically Link Existing Account::
+  Automatically link brokered identities without any validation with this authenticator. This is useful in an intranet environment of multiple user databases each with overlapping usernames/email addresses, but different passwords and you want to allow users to use any password without having to validate.  This is only reasonable if you manage all internal databases, and usernames/email addresses from one database matching those in another database belong to the same person.  Set the authenticator requirement to "Alternative".
+  
+NOTE: The described setup uses two authenticators, and is the simplest one, but it is possible to use other 
+authenticators according to your needs. For example you can add the `Review Profile` authenticator to the beginning of the flow if you still want end users to confirm their profile information.
```
matthewhelmke: ==== Automatically Link Existing First Login Flow
WARNING: The AutoLink authenticator would be dangerous in a generic environment where users can register themselves using arbitrary usernames/email addresses. Do not use this authenticator unless registration of users is carefully curated and usernames/email addresses are assigned, not requested. 

In order to configure a first login flow in which users are automatically linked without being prompted, create a new flow with the following two authenticators:

Create User If Unique::
This authenticator ensures unique users are handled. Set the authenticator requirement to "Alternative".

Automatically Link Existing Account::
Automatically link brokered identities without any validation with this authenticator. This is useful in an intranet environment of multiple user databases each with overlapping usernames/email addresses, but different passwords, and you want to allow users to use any password without having to validate. This is only reasonable if you manage all internal databases, and usernames/email addresses from one database matching those in another database belong to the same person. Set the authenticator requirement to "Alternative".
  
NOTE: The described setup uses two authenticators, and is the simplest one, but it is possible to use other 
authenticators according to your needs. For example, you can add the `Review Profile` authenticator to the beginning of the flow if you still want end users to confirm their profile information.

matthewhelmke (changes requested):
This is pretty good. I did a little cleaning up of the language in a couple places and the punctuation. I have my complete suggestion in the comment, so you can copy/paste over what you have and use my suggestion easily.

slominskir:
Copied and pasted then committed suggested edits in less than a minute.  Then spent and hour figuring out how to clean up line endings as I performed the copy and paste on a Windows box and it added CRLF line endings to all lines!  I think it's okay now.

matthewhelmke (approved):
Thank you, the changes look good. Sorry for the copy/paste Windows chaos. That had to be frustrating.

mposolda (approved):

PR #443, opened 2018-08-10
KEYCLOAK-8029 Documentation for Kerberos cross-realm trust

mposolda:


server_admin/topics/authentication/kerberos.adoc
```diff
@@ -188,6 +188,48 @@ WARNING: Credential delegation has some security implications so only use it if
          It's highly recommended to use it together with HTTPS.
          See for example http://www.microhowto.info/howto/configure_firefox_to_authenticate_using_spnego_and_kerberos.html[this article] for more details.
 
+==== Cross-realm trust
+
+In the Kerberos V5 protocol, the `realm` is a set of Kerberos principals defined in the Kerberos database (typically LDAP servers).
+Kerberos protocol has a concept of cross-realm trust. For example if there are 2 kerberos realms A and B, the cross-realm trust
+will allow the users from realm A to access resources (services) of realm B. This means that realm B trusts the realm A.
```
matthewhelmke: In the Kerberos V5 protocol, the `realm` is a set of Kerberos principals defined in the Kerberos database (typically LDAP servers).
The Kerberos protocol has a concept of cross-realm trust. For example, if there are 2 kerberos realms A and B, the cross-realm trust
will allow the users from realm A to access resources (services) of realm B. This means that realm B trusts the realm A.

server_admin/topics/authentication/kerberos.adoc
```diff
@@ -188,6 +188,48 @@ WARNING: Credential delegation has some security implications so only use it if
...
+
+.Kerberos cross-realm trust
+image:images/kerberos-trust-basic.png[]
+
+
+The {project_name} server has the support for cross-realm trust. There are few things which need to be done to achieve this:
```
matthewhelmke: No "the" is needed in "...server has support for..."

server_admin/topics/authentication/kerberos.adoc
```diff
@@ -188,6 +188,48 @@ WARNING: Credential delegation has some security implications so only use it if
...
+In general, it is needed to add the Kerberos principal `krbtgt/B@A` to both Kerberos databases of realm A and B. It is needed that
+this principal has same keys on both Kerberos realms. This is usually achieved when the principals have same password, key version number
+and there are same ciphers used in both realms. It is recommended to consult the Kerberos server documentation for more details.
+
+NOTE: The cross-realm is unidirectional by default. If you want it bidirectional and support that realm A trusts realm B,
+you need to add also the principal `krbtgt/A@B` to both Kerberos databases.
+However trust is transitive by default. If realm B trusts realm A and realm C trusts realm B, then realm C automatically trusts
+realm A as well without a need to have principal `krbtgt/C@A` available. Some additional configuration (for example `capaths`)
+may be needed to configure on Kerberos client side, so that the clients are able to find the trust path. Again, consult Kerberos
+documentation for more details.
```
matthewhelmke: NOTE: The cross-realm trust is unidirectional by default. If you want bidirectional trust to have realm A also trust realm B,
you must also add the principal `krbtgt/A@B` to both Kerberos databases. However, trust is transitive by default. If realm B trusts realm A and realm C trusts realm B, then realm C automatically trusts realm A without a need to have principal `krbtgt/C@A` available. Some additional configuration (for example `capaths`) may be needed to configure on Kerberos client side, so that the clients are able to find the trust path. Consult the Kerberos documentation for more details.

server_admin/topics/authentication/kerberos.adoc
```diff
@@ -188,6 +188,48 @@ WARNING: Credential delegation has some security implications so only use it if
...
+
+* Configure {project_name} server
+
+** If you use LDAP storage provider with Kerberos support, you need to configure server principal for realm B like for
+example `HTTP/mydomain.com@B` . The LDAP server must be able to find the users from realm A if you want users from realm A to
+successfully authenticate to {project_name}, as {project_name} server must be able to do SPNEGO flow and then find the users.
+So for example user kerberos principal `john@A` must be available as user in the LDAP under some LDAP DN
+like `uid=john,ou=People,dc=example,dc=com`. If you want both users from realm A and B to authenticate, you need to ensure that
+LDAP is able to find users from both realms A and B. We want to improve this limitation in future versions, so you can potentially
+create more separate LDAP providers for separate realms and ensure that SPNEGO works for both of them.
```
matthewhelmke: ** If you use an LDAP storage provider with Kerberos support, you need to configure the server principal for realm B as in this example: `HTTP/mydomain.com@B`. The LDAP server must be able to find the users from realm A if you want users from realm A to successfully authenticate to {project_name}, as {project_name} server must be able to do SPNEGO flow and then find the users.
For example, kerberos principal user `john@A` must be available as a user in the LDAP under an LDAP DN
such as `uid=john,ou=People,dc=example,dc=com`. If you want both users from realm A and B to authenticate, you need to ensure that LDAP is able to find users from both realms A and B. We want to improve this limitation in future versions, so you can potentially create more separate LDAP providers for separate realms and ensure that SPNEGO works for both of them.

server_admin/topics/authentication/kerberos.adoc
```diff
@@ -188,6 +188,48 @@ WARNING: Credential delegation has some security implications so only use it if
...
+
+** If you use Kerberos user storage provider (typically the Kerberos without LDAP integration), you need to configure the
+server principal as `HTTP/mydomain.com@B` and users from both Kerberos realms A and B should be able to authenticate.
```
matthewhelmke: Should be "** If you use a Kerberos user storage..."

server_admin/topics/authentication/kerberos.adoc
```diff
@@ -188,6 +188,48 @@ WARNING: Credential delegation has some security implications so only use it if
...
+
+WARNING: For Kerberos user storage provider, it is recommended that there are no conflicting users among kerberos realms.
+If they are, they will be mapped to same {project_name} user. This is again something, which we want to improve in future versions and
+provide some more flexible mappings from Kerberos principals to {project_name} usernames.
```
matthewhelmke: WARNING: For the Kerberos user storage provider, it is recommended that there are no conflicting users among kerberos realms. If conflicting users exist, they will be mapped to the same {project_name} user. This is also something, which we want to improve in future versions and provide some more flexible mappings from Kerberos principals to {project_name} usernames.

matthewhelmke (changes requested):
I have some edits for you, but this was mostly clear and easy to follow, I just cleaned up the language.

mposolda:
@matthewhelmke Updated per your suggestions. Thanks!

matthewhelmke (approved):

PR #450, opened 2018-08-20
KEYCLOAK-7719 Fixed endpoints auth/token list and also added well-kno…

stianst:
…wn endpoint

### DIFF
--- a/server_admin/topics/sso-protocols/oidc.adoc
+++ b/server_admin/topics/sso-protocols/oidc.adoc
@@ -95,10 +95,12 @@ Here's a list of OIDC endpoints that the {project_name} publishes.  These URLs a
 talk OIDC with the auth server.  These are all relative URLs and the root of the URL being the HTTP(S) protocol, hostname, and usually path prefixed with
 _/auth_:  i.e. $$https://localhost:8080/auth$$
 
-/realms/{realm-name}/protocol/openid-connect/token::
+/realms/{realm-name}/.well-known/openid-configuration::
+  This is the URL endpoint for the OpenID Connect Discovery endpoint.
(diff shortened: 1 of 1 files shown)


server_admin/topics/sso-protocols/oidc.adoc
```diff
@@ -95,10 +95,12 @@ Here's a list of OIDC endpoints that the {project_name} publishes.  These URLs a
 talk OIDC with the auth server.  These are all relative URLs and the root of the URL being the HTTP(S) protocol, hostname, and usually path prefixed with
 _/auth_:  i.e. $$https://localhost:8080/auth$$
 
-/realms/{realm-name}/protocol/openid-connect/token::
+/realms/{realm-name}/.well-known/openid-configuration::
+  This is the URL endpoint for the OpenID Connect Discovery endpoint.
```
matthewhelmke, line 99: It feels off to me to have the word "endpoint" repeated in the sentence, in both the subject and the definition. I think the following could work, or perhaps finding a different word or phrase for the end of the sentence.

This is the URL endpoint for OpenID Connect Discovery.

matthewhelmke (changes requested):
I left one comment.

server_admin/topics/sso-protocols/oidc.adoc
stianst: I didn't particularly like the way it's written either, but here I'm focusing on fixing the bug, not general edit of the page.

matthewhelmke: Okay, but since that sentence is new, I would suggest using the clearest phrasing possible to prevent future debt, when possible.

PR #456, opened 2018-08-29
KEYCLOAK-7437 Support for prompt=consent

mposolda:


securing_apps/topics/oidc/java/params_forwarding.adoc
```diff
@@ -23,6 +23,8 @@ enter the scope options `address phone`, then the request to {project_name} will
 
 * prompt - {project_name} supports these settings:
 ** `login` - SSO will be ignored and the {project_name} login page will be always shown, even if the user is already authenticated
+** `consent` - Applicable only for the clients with `Consent Required`. If it is used, the Consent page will be always displayed
+even if user already granted consent before to this client.
```
matthewhelmke: I suggest the following for the second sentence:
If it is used, the Consent p[age will always be displayed, even if the user previously granted consent to this client.

securing_apps/topics/oidc/javascript-adapter.adoc
```diff
@@ -317,7 +317,10 @@ Redirects to login form on (options is an optional object with redirectUri and/o
 Options is an Object, where:
 
 * redirectUri - Specifies the uri to redirect to after login.
-* prompt - By default the login screen is displayed if the user is not logged-in to {project_name}. To only authenticate to the application if the user is already logged-in and not display the login page if the user is not logged-in, set this option to `none`. To always require re-authentication and ignore SSO, set this option to `login` .
+* prompt - By default the login screen is displayed if the user is not logged-in to {project_name}. To only authenticate to the
+application if the user is already logged-in and not display the login page if the user is not logged-in, set this option to `none`.
+To always require re-authentication and ignore SSO, set this option to `login` .
+To always show the consent page even if user already granted consent to this client application, set this option to `consent` (Applicable only for the clients with `Consent Required`) .
```
matthewhelmke: Suggested edits:

* prompt - By default, the login screen is displayed if the user is not logged-in to {project_name}.
Set this option to `none` to authenticate to the application if the user is already logged-in and to not display the login page if the user is not logged-in.
Set this option to `login` to always require re-authentication and ignore current logged-in status and SSO.
Set this option to `consent` to always show the consent page, even if user already granted consent to this client application (this is applicable only for the clients with `Consent Required` set).

matthewhelmke (changes requested):
I am suggesting some changes. Please read them before applying to make sure I did not change the meaning unintentionally.

mposolda:
@matthewhelmke Thanks for the review! I've addressed per your feedback in the params_forwarding.adoc chapter. In the javascript_adapter.adoc section, I wasn't so sure and in the end, I've changed this section to add the link with more details to the `params_forwarding.adoc` where is the detailed description of the `prompt` parameter and all the possible values. That way, we don't duplicate docs for the explanation of `prompt` parameter in both places. Also for the future, we will need to edit just a single place (as the OIDC specification mentions some more values for the `prompt` parameter, which we don't yet support, but maybe we will support them in the future). WDYT?

matthewhelmke:
@mposolda I really like that idea. Having a single source and linking to it is a good plan.

matthewhelmke (approved):

mposolda:
@matthewhelmke Thanks!

PR #458, opened 2018-08-30
KEYCLOAK-8139 Update the documentation for supported/deprecated Wildf…

mposolda:
…ly/EAP adapters

### DIFF
--- a/securing_apps/topics/oidc/java/jboss-adapter.adoc
+++ b/securing_apps/topics/oidc/java/jboss-adapter.adoc
@@ -26,10 +26,13 @@ Both methods are described in this section.
 
 Adapters are available as a separate archive depending on what server version you are using.
 
-NOTE: {appserver_name} should be running when you install the adapter. If you have either one running, you must stop it before installing and then restart it after installation is complete.
-
 ifeval::[{project_community}==true]
-Install on WildFly 9, 10 or 11:
+
+NOTE: We only test and maintain adapter with the most recent version of WildFly available upon the release. Once new version of
+WildFly is released, the current adapters become deprecated and support for them will be removed after next WildFly release.
+The other alternative is to switch your applications from WildFly to the JBoss EAP, as the JBoss EAP adapter is supported for much longer period.
+
+Install on WildFly 9 or newer:
 
 [source, subs="attributes"]
 ----
@@ -105,7 +108,7 @@ To configure the adapter subsystem if the server is not running execute:
 NOTE: Alternatively, you can specify the `server.config` property while installing adapters from the command line to install adapters using a different config, for example: `-Dserver.config=standalone-ha.xml`.
 
 ifeval::[{project_community}==true]
-.WildFly 11
+.WildFly 11 or newer
 [source]
 ----
 $ ./bin/jboss-cli.sh --file=adapter-elytron-install-offline.cli
@@ -116,10 +119,14 @@ $ ./bin/jboss-cli.sh --file=adapter-elytron-install-offline.cli
 ----
 $ ./bin/jboss-cli.sh --file=adapter-install-offline.cli
 ----
+
+NOTE: It is possible to use the legacy non-Elytron adapter on WildFly 11 or newer as well, meaning you can use `adapter-install-offline.cli`
+even on those versions. However, we recommend to use the newer Elytron adapter.
+
 endif::[]
 
 ifeval::[{project_product}==true]
-.JBoss EAP 7.1
+.JBoss EAP 7.1 or newer
 [source]
 ----
 $ ./bin/jboss-cli.sh --file=adapter-elytron-install-offline.cli
@@ -130,15 +137,18 @@ $ ./bin/jboss-cli.sh --file=adapter-elytron-install-offline.cli
 ----
 $ ./bin/jboss-cli.sh --file=adapter-install-offline.cli
 ----
-endif::[]
 
+NOTE: It is possible to use the legacy non-Elytron adapter on JBoss EAP 7.1 or newer as well, meaning you can use `adapter-install-offline.cli`
+even on those versions. However, we recommend to use the newer Elytron adapter.
 
 NOTE: The offline script is not available for JBoss EAP 6.4
+
+endif::[]
 
 Alternatively, if the server is running execute:
 
 ifeval::[{project_community}==true]
-.WildFly 11
+.WildFly 11 or newer
 [source]
 ----
 $ ./bin/jboss-cli.sh --file=adapter-elytron-install.cli
@@ -152,7 +162,7 @@ $ ./bin/jboss-cli.sh --file=adapter-install.cli
 endif::[]
 
 ifeval::[{project_product}==true]
-.JBoss EAP 7.1
+.JBoss EAP 7.1 or newer
 [source]
 ----
 $ ./bin/jboss-cli.sh --file=adapter-elytron-install.cli
--- a/securing_apps/topics/saml/java/jboss-adapter/jboss_adapter_installation.adoc
+++ b/securing_apps/topics/saml/java/jboss-adapter/jboss_adapter_installation.adoc
@@ -5,7 +5,12 @@
 Each adapter is a separate download on the {project_name} download site.
 
 ifeval::[{project_community}==true]
-Install on WildFly 9 or 10, 11 or JBoss EAP 7:
+
+NOTE: We only test and maintain adapter with the most recent version of WildFly available upon the release. Once new version of
+WildFly is released, the current adapters become deprecated and support for them will be removed after next WildFly release.
+The other alternative is to switch your applications from WildFly to the JBoss EAP, as the JBoss EAP adapter is supported for much longer period.
+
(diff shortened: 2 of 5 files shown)


securing_apps/topics/oidc/java/jboss-adapter.adoc
```diff
@@ -116,10 +120,14 @@ $ ./bin/jboss-cli.sh --file=adapter-elytron-install-offline.cli
 ----
 $ ./bin/jboss-cli.sh --file=adapter-install-offline.cli
 ----
+
+NOTE: It is possible to use legacy non-elytron adapter on Wildfly 11 or newer as well, hence you can use `adapter-install-offline.cli`
+even on those versions. However we recommend to use newer Elytron adapter.
```
matthewhelmke: I have a couple minor changes to suggest here:

NOTE: It is possible to use the legacy non-Elytron adapter on Wildfly 11 or newer as well, meaning you can use `adapter-install-offline.cli` even on those versions. However, we recommend to use the newer Elytron adapter.

securing_apps/topics/oidc/java/jboss-adapter.adoc
```diff
@@ -130,15 +138,18 @@ $ ./bin/jboss-cli.sh --file=adapter-elytron-install-offline.cli
 ----
 $ ./bin/jboss-cli.sh --file=adapter-install-offline.cli
 ----
-endif::[]
 
+NOTE: It is possible to use legacy non-elytron adapter on JBoss EAP 7.1 or newer as well, hence you can use `adapter-install-offline.cli`
```
matthewhelmke: same change as above

securing_apps/topics/saml/java/jboss-adapter/jboss_adapter_installation.adoc
```diff
@@ -63,10 +67,14 @@ $ ./bin/jboss-cli.sh --file=adapter-elytron-install-saml.cli
 $ cd $JBOSS_HOME/bin
 $ jboss-cli.sh -c --file=adapter-install-saml.cli
 ----
+
+NOTE: It is possible to use legacy non-elytron adapter on Wildfly 11 or newer as well, hence you can use `adapter-install-saml.cli`
+even on those versions. However we recommend to use newer Elytron adapter.
+
```
matthewhelmke: same note as previous.

securing_apps/topics/saml/java/jboss-adapter/jboss_adapter_installation.adoc
```diff
@@ -79,6 +87,10 @@ $ ./bin/jboss-cli.sh --file=adapter-elytron-install-saml.cli
 $ cd $JBOSS_HOME/bin
 $ jboss-cli.sh -c --file=adapter-install-saml.cli
 ----
+
+NOTE: It is possible to use legacy non-elytron adapter on JBoss EAP 7.1 or newer as well, hence you can use `adapter-install-saml.cli`
+even on those versions. However we recommend to use newer Elytron adapter.
+
```
matthewhelmke: same note as before. I wonder whether we should pull this note out into a separate file and then include it in all these locations or whether that is more work than remembering to update it in four places in the future. I would lean toward pulling it out and including it, so that we only update in one place in the future.

matthewhelmke (changes requested):

securing_apps/topics/oidc/java/jboss-adapter.adoc
```diff
@@ -29,7 +29,11 @@ Adapters are available as a separate archive depending on what server version yo
 NOTE: {appserver_name} should be running when you install the adapter. If you have either one running, you must stop it before installing and then restart it after installation is complete.
 
 ifeval::[{project_community}==true]
-Install on WildFly 9, 10 or 11:
+
+NOTE: WildFly adapters on version 11 or older are deprecated. They may still work, but we cannot guarantee to regularly test them
```
hmlnarik: _I'd rather not mention version number explicitly. WDYT about something about the following formulation?_

We only test and maintain adapter with the most recent version of WildlFly available upon the release. Once new version of WildFly is released, the current adapters become deprecated and support for them will be removed after next WildFly release.

securing_apps/topics/oidc/java/jboss-adapter.adoc
```diff
@@ -29,7 +29,11 @@ Adapters are available as a separate archive depending on what server version yo
 NOTE: {appserver_name} should be running when you install the adapter. If you have either one running, you must stop it before installing and then restart it after installation is complete.
```
vramik: This note seems to be outdated. It applies only to installing adapters in online mode. Since WF11 there is possibility to install those in offline mode.

securing_apps/topics/oidc/java/jboss-adapter.adoc
matthewhelmke: I like that!

mposolda:
@matthewhelmke @hmlnarik @vramik Thanks everyone for the feedback. I've incorporated the changes in separate commit of this PR just for the easier review: https://github.com/keycloak/keycloak-documentation/pull/458/commits/8411b34fddc83e340c027a6cb3c386391b5dc48f

Once we agree on the content, I will squash the 2 commits into the single one. What I did was:

- Removed the outdated NOTE mentioned by Vlasta

- Changed the paragraph about latest supported WildFly release as proposed by Hynek. Agree it's better to remove the versions. I've just added one more sentence to it to emphasize that people have still an alternative to move to JBoss EAP. This one: 
````
The other alternative is to switch your applications from WildFly to the JBoss EAP, as the JBoss EAP adapter is supported for much longer period.
````

- Changed the paragraph about non-elytron adapter as proposed by Matthew. In the end, I've just changed it, but didn't introduce any new "single" place for this content. It is quite short paragraph, there are 2 variables in it (server name + file name), which would require 2 macros and the content shouldn't change in next versions (hopefully) as Wildfly 11 and JBoss EAP 7.1 will always be the "first versions" where the elytron adapter was available, so formulation like "Wildfly 11 or newer" will be always valid even in future versions.

File: securing_apps/topics/oidc/java/jboss-adapter.adoc [hunk 7]
```diff
@@ -26,10 +26,13 @@ Both methods are described in this section.
...
-NOTE: {appserver_name} should be running when you install the adapter. If you have either one running, you must stop it before installing and then restart it after installation is complete.
-
 ifeval::[{project_community}==true]
-Install on WildFly 9, 10 or 11:
+
+NOTE: We only test and maintain adapter with the most recent version of WildFly available upon the release. Once new version of
+WildFly is released, the current adapters become deprecated and support for them will be removed after next WildFly release.
+The other alternative is to switch your applications from WildFly to the JBoss EAP, as the JBoss EAP adapter is supported for much longer period.
+
+Install on WildFly 9, 10, 11 or 13:
```
hmlnarik: Install on WildFly 9 _or newer_?

mposolda:
@hmlnarik +1 for "Wildfly 9 or newer" . I've changed in both OIDC and SAML documentation, so hopefully for next WildFly upgrade, we don't need to touch adapter docs as exact versions are not here anymore :)

Among this, I've added the release notes about the WildFly 13 upgrade and changed the infinispan-server version in cross-dc documentation to 9.2.4, which is the version we're testing with (I think I've already changed that in the previous PR for Wildfly 13 documentation, but probably was lost during rebase...).

vramik:
with 

> WildFly 9 or newer

it's easier. but it also contains not supported wf12. Is there somewhere mentioned we doesn't support wf12?

mposolda:
Not sure. In the previous paragraph, it is mentioned that Wildfly adapters for older versions are deprecated. Also the downloads page don't specify the Wildfly 12 (where the versions 9,10,11 are specified, even if they are deprecated). Isn't this sufficient? What we can do is:

- Keep as it is in my last commit (that's my preference)

- Revert back and explicitly mention "Wildfly 9,10,11 or 13" as was before Hynek suggested to change it

- Keep as it is in my last commit, but add some additional WARNING explicitly about Wildfly 12 adapter and mention that it's highly unrecommended due the bug in the WildFly.

@hmlnarik @vramik What do you think?

vramik (dismissed):
@mposolda after going through the note again

> We only test and maintain adapter with the most recent version of WildFly available upon the release. Once new version of WildFly is released, the current adapters become deprecated and support for them will be removed after next WildFly release.

it doesn't applies to me we support last two versions of wildlfy (which was my main concern, because it'd contain the wf12). So I think we should keep it as it is now.

hmlnarik (commented):
LGTM, thank you!

I believe the second commit is for easier review of the changes, which I find beneficial. I'd approve it after @matthewhelmke's approval and squashing it to a single commit.

mposolda:
@hmlnarik Thanks! Yes, the second commit is just for easier review of the changes.

matthewhelmke (dismissed):

hmlnarik (approved):
Thanks for squashing, approving and merging based on previous approvals.

PR #459, opened 2018-08-30
Patch about auth and token endpoint

jolisylvain:
I think the definitions of auth and token OIDC endpoints have been inverted


matthewhelmke (approved):
Thank you! This looks fine to me, but I will defer to engineering for confirmation.

stianst:
Actually, it's both inverted and also wrong. The auth endpoint is for the authorization code flow that's correct, but not for obtaining tokens via implicit flow, direct grant or client grants. All requests that obtain tokens use the token endpoint.

stianst (changes requested):
See comment

stianst:
Closing this as it's incorrect and there is no feedback from contributor. Feel free to re-open once comments have been addressed.

PR #460, opened 2018-08-31
Update boot.adoc

jkenefick:
.../bin/standalone.sh didn't work for me when running on a bash terminal on a mac.  This should work for everyone.

### DIFF
--- a/getting_started/topics/first-boot/boot.adoc
+++ b/getting_started/topics/first-boot/boot.adoc
@@ -6,7 +6,8 @@ To boot the {project_name} server, go to the `bin` directory of the server distr
 .Linux/Unix
 [source,bash,subs=+attributes]
 ----
-$ .../bin/standalone.sh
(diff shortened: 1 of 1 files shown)


matthewhelmke:
Thank you.

I see what you are aiming for here and I agree that we can do something to make this better. The commands as you adjusted them worked for you because you knew that `cd bin` was referring to the location within the server distribution, and not to the /bin directory in the file system hierarchy. I don't know if everyone will catch that right away and we may be exchanging one less-than-ideal step for another.

Let me give this further thought and see if I can come up with something even clearer.

matthewhelmke (approved):
After giving this some thought (and remembering that the section is introduced with "To boot the {project_name} server, go to the `bin` directory of the server distribution and run the `standalone` boot script:") I have reconsidered. I think your solution is a good one. Sorry for the delay.

PR #461, opened 2018-08-31
Update before.adoc

jkenefick:
I was confused when I saw that wildfly also ran as part of keycloak and that I was able to deploy to it.  Then afterwards I realised that I should have downloaded a separate instance, but then I was stuck with two vanillas, one on the keycloak wildfly and one on the downloaded wildfly.  Putting in an acknowledgement of the existence of wildfly on the keycloak installation might help alleviate confusion.


matthewhelmke (approved):
Thank you. This is a significant improvement in clarity.
We always have two reviewers: me from a technical documentation perspective, and a review from someone in engineering. After both reviews are complete, we are ready to merge.

PR #462, opened 2018-08-31
Update before.adoc

jkenefick:


matthewhelmke:
Thank you.
I'll make the same comment here as in this PR: https://github.com/keycloak/keycloak-documentation/pull/460#issuecomment-417713798

matthewhelmke (approved):
After giving this some thought (and remembering that the section is introduced with "To boot the {project_name} server, go to the `bin` directory of the server distribution and run the `standalone` boot script:") I have reconsidered. I think your solution is a good one. Sorry for the delay.

PR #463, opened 2018-08-31
Update install-client-adapter.adoc

jkenefick:
To make the user's life a little easier.

### DIFF
--- a/getting_started/topics/secure-jboss-app/install-client-adapter.adoc
+++ b/getting_started/topics/secure-jboss-app/install-client-adapter.adoc
@@ -1,7 +1,7 @@
 
 === Installing the Client Adapter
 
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
Thank you. This looks good to me. Someone from engineering also needs to review and approve, to make sure the change is appropriate from their perspective.

stianst:
I'm going to reject this as it doesn't work for RH-SSO builds of the documentation where we're talking about JBoss EAP and not WildFly, hence the app container variable. Maintaining both sets of links would be relatively complex especially since there's no easy direct link to download JBoss EAP. I also don't believe this change is that important as I assume if you want to use the WildFly adapter you already know where to get it from and most likely already have it installed.

PR #464, opened 2018-08-31
Update install-client-adapter.adoc

jkenefick:
Took a small bit of time on my computer.


matthewhelmke (approved):
Thank you. Looks like a useful change to me.

We still need a second review from someone in engineering before merging.

PR #466, opened 2018-09-04
[KEYCLOAK-6962] Added ways to (minor) upgrade rpm distribution

pskopek:
https://issues.jboss.org/browse/KEYCLOAK-6962

### DIFF
--- a/upgrading/topics/install_new_version.adoc
+++ b/upgrading/topics/install_new_version.adoc
@@ -18,18 +18,54 @@ in the new installation.
 +
 For domain installations, create the empty directory {project_dirref}/domain/deployments.
 +
-NOTE: 
+NOTE:
 Files in the bin directory should not be overwritten by the files from previous versions. Changes should be made manually.
 
 . Copy any custom modules that have been added to the modules directory.
 . Run the applicable upgrade script below.
+
+ifeval::[{project_product}==true]
+To upgrade {project_name} server RPM distribution, complete the following steps:
+
+. Prior to applying the upgrade, handle any open transactions and delete the /var/opt/rh/rh-sso7/lib/keycloak/standalone/data/tx-object-store/ transaction directory.
+. Make sure you have subscribed to proper repositories containing JBOSS EAP and {project_name}.
+
+ subscription-manager repos --enable=jb-eap-7.1-for-rhel-7-server-rpms
+ subscription-manager repos --enable=rh-sso-7.2-for-rhel-7-server-rpms
++
+[NOTE]
+====
+To disable older product repositories for both JBOSS EAP and {project_name} use:
+
+ subscription-manager repos --disable=<OLDER_PRODUCT_REPO>
+
+To check the repositories use:
+
+ yum repolist
+====
+. The RPM upgrade process will not replace any of your modified configuration files, and will instead create .rpmnew files for the default configuration of the new {project_name} version.
++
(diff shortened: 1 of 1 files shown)


matthewhelmke:
We have information in https://github.com/keycloak/keycloak-documentation/blob/master/upgrading/topics/rhsso/patching-rpm-installation.adoc for updating via RPMs. This topic is RH-SSO only, as Keycloak is not distributed/updated using RPMs. It looks like what you have written may overlap and expand upon what we already have here. Can you compare the two and see how to merge the information?

pskopek:
@matthewhelmke the topic you mentioned is for micro upgrade (or patching). There is no need to do anything with already subscribed repositories and it is correct. 
The one from this section describes minor upgrade (7.1 -> 7.2).

matthewhelmke (dismissed):
Got it. Thank you, @pskopek . This looks good.

hmlnarik (changes requested):
Good point @matthewhelmke, we have to enclose the RPM-related stuff into `ifeval::[{project_product}==true]` similarly to https://github.com/keycloak/keycloak-documentation/blob/0e8b9a7ac2207db7b4dedb8be95413c7a18e6eb7/upgrading/topics.adoc.

In the meantime, @wyvie could you please have a look at this PR?

pskopek:
@hmlnarik good catch, fixed

upgrading/topics/install_new_version.adoc
```diff
@@ -18,18 +18,48 @@ in the new installation.
...
+To upgrade {project_name} server RPM distribution, complete the following steps:
+
+. Prior to applying the upgrade, handle any open transactions and delete the /var/opt/rh/rh-sso7/lib/keycloak/standalone/data/tx-object-store/ transaction directory.
+. Make sure you have subscribed to proper repositories containing JBOSS EAP and {project_name}.
+
+ subscription-manager repos --enable=jb-eap-7.1-for-rhel-7-server-rpms
+ subscription-manager repos --enable=rh-sso-7.2-for-rhel-7-server-rpms
++
+NOTE:
+ Use to disable older product repositories for both JBOSS EAP and {project_name} use:
```
alicerum: two times 'use' here. First one could possibly be removed

alicerum (commented):
otherwise all seems good

pskopek:
Thnks @wyvie, fixed.

upgrading/topics/install_new_version.adoc
```diff
@@ -18,18 +18,48 @@ in the new installation.
...
+
+ subscription-manager repos --enable=jb-eap-7.1-for-rhel-7-server-rpms
+ subscription-manager repos --enable=rh-sso-7.2-for-rhel-7-server-rpms
++
+NOTE:
+ To disable older product repositories for both JBOSS EAP and {project_name} use:
+
+ subscription-manager repos --disable=<OLDER_PRODUCT_REPO>
++
+NOTE:
```
hmlnarik: This does not render correctly to an info box. The `NOTE: ` has to be followed by text, otherwise it renders just as "NOTE: " text.

hmlnarik (changes requested):

hmlnarik (approved):

hmlnarik:
@pskopek Thanks for changes, @matthewhelmke, @wyvie Thanks for review!

PR #472, opened 2018-09-07
KEYCLOAK-8159 return missing section

matthewhelmke:
This returns the "Advanced Concepts" section.

When approved, this will also need to be cherry-picked back to the CD03 branch as the content is missing from there.


iankko:
@matthewhelmke LGTM, ACK. @stianst PTAL 

Thank you, Jan

stianst (approved):

matthewhelmke:
After some discussion, it was decided not to cherry-pick this back to CD03, but the fix is merged and will be included in CD04.

PR #476, opened 2018-09-13
KEYCLOAK-7612 add note that jpa-entity-provider is unsupported

matthewhelmke:


matthewhelmke:
Hmm. I thought all the openid.net URLs were updated in #470, but this build also failed.

abstractj (approved):

PR #477, opened 2018-09-13
typo: complete to completely

bhamail:


matthewhelmke (approved):
Nice catch, thank you!

We always have two people look at PRs, me from a documentation/writing perspective, and also someone from engineering. Sometimes that can take a while, but no worries. We tend to merge a bunch of PRs at the same time, shortly before a release. This one is a no-brainer and much appreciated! Thanks, again!

PR #494, opened 2018-10-10
[KEYCLOAK-8449] - Option to automatically map HTTP verbs to scopes when configuring the policy enforcer

pedroigor:


matthewhelmke (approved):
LGTM

PR #504, opened 2018-10-19
[KEYCLOAK-4342] - Make adapter state cookie path configurable

pedroigor:


matthewhelmke (dismissed):
LGTM

pedroigor:
@matthewhelmke, upstream merged. Should I restart a build ? Errors seems unrelated with these changes ...

matthewhelmke:
Yes

pedroigor:
@matthewhelmke, I was able to fix some of errors. Are they:

* Invalid external links in gatekeeper doc. We are ignoring some of them as they are just examples
* Invalid external link in token exchange
* Additional check to anchors in order to also match the anchor within `href` attributes

I hope we can get another gatekeeper error fixed (external link) when the image is deployed to docker hub.

Remaining issues are references to internal anchors, which I'm really not sure why is failing. I suspect, but could not find an easy fix.

Will merge as is.

matthewhelmke (dismissed):
Thank  you, @pedroigor

pedroigor:
Remaining errors seems to be related with gatekeeper docs, probably because of some links pointing to rls that will be there after a release.

matthewhelmke:
Thanks for looking into all the errors.

PR #506, opened 2018-10-25
KEYCLOAK-8175 Client scopes permissions

mposolda:


mposolda:
Codebase PR: https://github.com/keycloak/keycloak/pull/5670

server_admin/topics/clients/client-scopes.adoc
```diff
@@ -131,6 +131,17 @@ To see an example of a real access token, generated for the particular user and
...
+
+When issuing tokens for the particular user, the client scope is applied just if the user is permitted to use it. In case that
+client scope does not have any role scope mappings defined on itself, then each user is automatically permitted to use this client scope.
+However in case that client scope has any role scope mappings defined on itself, then user must be member of at least one of those roles.
+In other words, there must be any intersection among the user roles and the roles of the client scope. Composite roles are taken into
+account when evaluating this intersection.
+
+If user is not permitted to use the client scope, then any protocol mappers or role scope mappings won't be used when generating tokens.
+Also the client scope won't appear in the _scope_ value in the token.
+
```
matthewhelmke, line 144: When issuing tokens for a particular user, the client scope is applied only if the user is permitted to use it. In the case that
a client scope does not have any role scope mappings defined on itself, then each user is automatically permitted to use this client scope. However, when a client scope has any role scope mappings defined on itself, then the user must be a member of at least one of the roles. In other words, there must be an intersection between the user roles and the roles of the client scope. Composite roles are taken into account when evaluating this intersection. If a user is not permitted to use the client scope, then no protocol mappers or role scope mappings will be used when generating tokens and the client scope will not appear in the _scope_ value in the token.

matthewhelmke (changes requested):
Again, very few corrections, but it was such a short passage that I copied/pasted/edited and pasted the new version into a comment.

mposolda:
@matthewhelmke Thanks! I've applied your suggestions. However I still kept two paragraphs. I guess you didn't mean to merge it into single paragraph, but it's just the github showing the comment in single paragraph? Let me know and I will merge into single paragraph if it's not the case and you think that single paragraph is better.

matthewhelmke:
@mposolda You are correct. I intended to leave the two paragraphs separate.

matthewhelmke (approved):

PR #507, opened 2018-10-26
KEYCLOAK-8485 Document priorities used by protocol mappers

mposolda:


server_admin/topics/clients/protocol-mappers.adoc
```diff
@@ -39,3 +39,16 @@ image:{project_images}/add-mapper.png[]
...
+Mapper implementations have _priority order_. This priority order is not the configuration property of the mapper, however it is
+the property of the concrete implementation of the mapper.
+
+You will see mappers sorted by the order in the list of mappers in the admin console and the changes in the token or assertion
+will be applied by that order - lowest go first. This allows to ensure that some implementations, which are dependent on the
+others, are processed afterwards.
+
+For example we first want to compute the roles, which will go to the token. Then we resolve audiences based on those roles. Finally
+we can eventually process some javascript script, which will have the roles and audiences already available in the token.
+
```
matthewhelmke, line 51: Mapper implementations have _priority order_. This priority order is not the configuration property of the mapper; rather, it is the property of the concrete implementation of the mapper.

Mappers are sorted in the admin console by the order in the list of mappers and the changes in the token or assertion will be applied using that order with the lowest being applied first. This means that implementations which are dependent on other implementations are processed in the needed order.

For example, when we first want to compute the roles which will be included with a token, we first resolve audiences based on those roles. Then, we process a JavaScript script that uses the roles and audiences already available in the token.

matthewhelmke (changes requested):
I again did a copy/paste/edit/post. This is a cool feature!

matthewhelmke (approved):

mposolda:
@matthewhelmke Thanks for your suggestions Matthew!

PR #508, opened 2018-10-26
[KEYCLOAK-8353] - Entitlement API class does not exist any more

pedroigor:


matthewhelmke:
Is there another commit coming, @pedroigor ? Since one of the images is being removed, I am thinking that surely there is a text file somewhere with a reference that must also be removed.

pedroigor:
@matthewhelmke, The image I'm removing does not have any reference to it. I think I left it behind when updating docs to new architecture based on UMA 2.

matthewhelmke (approved):
@pedroigor Thank you for the explanation. Looks good, then. :+1:

PR #516, opened 2018-11-14
KEYCLOAK-8482 Access token should never contain azp as an audience

mposolda:


server_admin/topics/clients/oidc/audience.adoc
```diff
@@ -98,6 +98,11 @@ on the `my-app` client, but instead create a dedicated client scope, for example
 for the client roles of the `good-service` client. Assuming that this client scope will be added as an optional client scope to
 the `my-app` client, the client roles and audience will be added to the token just if explicitly requested by the `scope=good-service` parameter.
 
+NOTE: The frontend client itself is not automatically added to the access token audience. This allows to easily differentiate access token and
+ID token as access token will not contain the client, for which the token was issued, as an audience. So in the example above, the `my-app`
+won't be added as an audience. If you really need client itself as an audience, then see below for the <<_audience_hardcoded, hardcoded audience>>
+option. However using same client as frontend and REST service is not recommended anyway.
```
matthewhelmke: NOTE: The frontend client itself is not automatically added to the access token audience. This allows for easy differentiation between the access token and the ID token, because the access token will not contain the client for which the token was issued as an audience. So in the example above, the `my-app` won't be added as an audience. If you need the client itself as an audience, see the <<_audience_hardcoded, hardcoded audience>> option. However, using the same client as both frontend and REST service is not recommended.

matthewhelmke (changes requested):
I have made some edits.

mposolda:
@matthewhelmke Updated as you suggested. Thanks!

matthewhelmke:
Thank you, @mposolda

matthewhelmke (approved):

PR #521, opened 2018-11-18
JS adapter promises have changed

gcleaves:
It appears that instead of `.then` and `.catch` you now use `.success` and `.error`. I've updated the examples accordingly.

### DIFF
--- a/securing_apps/topics/oidc/javascript-adapter.adoc
+++ b/securing_apps/topics/oidc/javascript-adapter.adoc
@@ -28,9 +28,9 @@ The following example shows how to initialize the JavaScript adapter:
     <script src="keycloak.js"></script>
     <script>
         var keycloak = Keycloak();
-        keycloak.init().then(function(authenticated) {
+        keycloak.init().success(function(authenticated) {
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
Thank you. This looks good to me. Engineering is super busy at the moment, but hopefully someone from their will also review soon.

sexygeoff:
Is there any way to apply these documentation changes into the most recently tagged release instead of master? Forgive my lack of understanding the inner workings of Github. But I think that if you make the improvements in master and these aren't pushed to your public documentation site, then everybody continues to read outdated docs.

matthewhelmke:
I definitely understand the desire to do that, @sexygeoff. I would like to be able to push changes back to already released documents, too. Here is what we can do:

We apply documentation changes to master, then they are pushed out at the next release. It is primarily a question of staffing and logistics. This is a straightforward method that requires some, but not extensive time, whereas pushing the changes from master into existing releases is a much more time-consuming process and we just don't have the ability to do it while keeping up with everything else.

PR #522, opened 2018-11-19
?html is not allowed when auto-escaping is on

jpenren:
Keycloak v.4.6.0
Caused by: freemarker.core.ParseException: Syntax error in template "account.ftl" in line 66, column 143:                           
Using ?html (legacy escaping) is not allowed when auto-escaping is on with a markup output format (HTML), to avoid double-escaping m
istakes

### DIFF
--- a/server_development/topics/custom-attributes.adoc
+++ b/server_development/topics/custom-attributes.adoc
@@ -41,7 +41,7 @@ As an example to add a mobile number to the account page add the following snipp
    </div>
 
    <div class="col-sm-10 col-md-10">
-       <input type="text" class="form-control" id="user.attributes.mobile" name="user.attributes.mobile" value="${(account.attributes.mobile!'')?html}"/>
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
Thanks for catching this and sharing an edit! We appreciate it. We have two reviews in our process, one from me that is writing-focused and one from engineering to check technical info. You just passed my review. Someone from engineering will take a look as soon as possible (this is a crazy week, so it may take longer than usual, but it will happen).

ssilvert (approved):
This is fine as-is for the docs.  But for any value that allows HTML, we use the keycloak sanitizer.   So we turn escaping off to make the HTML work, but we use the sanitizer to make sure that the HTML is not malicious.  Here is an example from src/main/resources/theme/base/account/template.ftl:

 ${kcSanitize(message.summary)?no_esc}

I'm thinking this should be mentioned somewhere in the docs.

matthewhelmke:
@ssilvert I have created a new Jira ticket for working this in to our docs. 
https://issues.jboss.org/browse/KEYCLOAK-8870

PR #531, opened 2018-11-28
Add additional examples for custom scripts on authentication flow

ThoTischner:
With this change end users are able to understand the custom execution scripts better for authentication flows.

If you want to deny access to applications on keycloak level via role or group membership, this will be working examples.

### DIFF
--- a/server_admin/topics/authentication/flows.adoc
+++ b/server_admin/topics/authentication/flows.adoc
@@ -97,4 +97,74 @@ function authenticate(context) {
   context.success();
 }
 ----                        
+
+More examples
+
+This script will do additional checks for clients that have "foobar" as client-id prefix.
+It will then check if the user is member of a group with the same name like the client-id to which the user tries to login.
+
+[source,javascript]
+----
+function authenticate(context) {
+
+LOG.info(script.name + ': started');
+
+var client_id = httpRequest.getUri().getQueryParameters().get("client_id").toString();
+client_id = client_id.substring(1, client_id.length() - 1);
+
+if ( client_id.indexOf("foobar") !== 0 ) {
+LOG.info(script.name + ": allow login for non foobar client " + client_id);
+context.success();
+return;
+}
+
+var group = realm.searchForGroupByName(client_id, 0, 1)[0];
+
+if ( user.isMemberOf(group) ) {
+LOG.info(script.name + ": allow login for foobar user " + user.username + 
+" for group " + group.getName());
+context.success();
+return;
+}
+
+LOG.info(script.name + ": deny login for " + user.username);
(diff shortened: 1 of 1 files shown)


matthewhelmke (approved):
Thank you very much! This looks good to me. We always have two reviews on docs contributions: one from me for writing and one from someone in engineering. You have passed my review.

abstractj:
@matthewhelmke @disappear89 even though I think is great to have some examples documented. I believe it's better to put them into the quickstarts. Code like this into the docs can become outdated, due to API changes and remain undetected. While if you contribute to one of our quickstarts (https://github.com/keycloak/keycloak-quickstarts), tests will break, code won't run.

Also, please make sure to have a Jira to each PR you submit, this help us to prioritize.

ThoTischner:
@matthewhelmke Thats a good idea but dont you think its a bit hard to write a quickstart with all that tests and so, maybe you could assist me?
I think event if there is a quickstart, we have still a lack on docs cause i found this solution only by googling on a mailing list archive..

matthewhelmke:
@disappear89 @abstractj 
I do not interact with the quickstarts, they are completely maintained by engineers, so I am not equipped to assist with those. I do know that they contain many code samples and tests that keep those code samples working and prevent them from being outdated, so there are some real benefits to trying to push this over to the quickstarts. @abstractj, do you have time or know who might have time to help with the process, since we have such a willing contributor? I do realize we may need to wait a few weeks as we are in a crunch time nearing the end of a big release cycle.

abstractj:
@disappear89 @matthewhelmke I believe a great starting point would be to create an enhancement to the quickstarts and the use case scenario for these examples. In this way we can plan the further releases. If you would like to look at JS related stuff, take a look at these examples: https://github.com/keycloak/keycloak-quickstarts/tree/latest/app-angular2 and https://github.com/keycloak/keycloak-quickstarts/tree/latest/app-jee-html5.

I'm aware that this path will takes more time and probably more effort. But I believe that it pays off in the long run.

ThoTischner:
@abstractj I totally agree that it would be better to have a solution with tests. But i really don`t know how to transfer your quick start links to a working solution for a custom authenticator flow with a script execution.
I dont see any comparable config on this quick starts. They focus on integrating and testing new clients or not?
We dont want to attach a client, we just found and want to document a generic solution that works with every client to limit allowed users via groups or roles on keycloak level, no matter which type of client is connected.

I think this will be close to what we need to do https://github.com/keycloak/keycloak-quickstarts/tree/latest/action-token-authenticator but this is also way more than we have implemented. We dont have a custom authenticator just a script as required execution on our browser login flow..

stianst:
Closing this as we should not have too many code examples in documentation. They are hard to maintain and not tested so are likely not to work.

PR #535, opened 2018-12-03
KEYCLOAK-8737 update release notes for RH-SSO 7.3

matthewhelmke:
@stianst this is so much simpler than what I was trying to do here: https://gitlab.cee.redhat.com/red-hat-single-sign-on-documentation/doc-release-notes/merge_requests/18/diffs

I am concerned that I am missing something. Specifically, where do we link to the supported component versions, fixed issues, and supported configurations (which is being done in the GitLab update above).

### DIFF
--- a/release_notes/master.adoc
+++ b/release_notes/master.adoc
@@ -8,6 +8,16 @@ include::topics/templates/document-attributes-product.adoc[]
 
 = {releasenotes_name}
 
+ifeval::[{project_product_cd}==false]
+
+== {project_name_full} 7.3
+
+include::topics/7_3_final.adoc[leveloffset=2]
+
+endif::[]
+
+ifeval::[{project_product_cd}==true]
+
 == {project_name_full} 7.3.CD05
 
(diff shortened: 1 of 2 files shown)


stianst (dismissed):
I'm afraid we're going to need a curated list for 7.3.0.GA rather than simply including changes from Keycloak. At the very least the list needs to be ordered by importance, probably with a separate section on the end for misc. There is also probably a number of things that can be combined into sensible sections. We also need John's input on reviewing.

stianst:
We also need to point out that Authorization Services are now fully supported, with the exceptino of the Drools Policy

matthewhelmke:
@stianst That actually feels better to me. I created a curated list in the PR (to the wrong repo). I'll revisit and revise and update here.

matthewhelmke:
I have an official "Looks good, thanks" from John, with one edit in the Gdoc.

matthewhelmke:
There is likely to be one more commit with final edits to the Authz section and with Known Issues included.

I can squash commits at that time.

stianst (approved):

PR #541, opened 2018-12-05
Update to EAP 7.2

stianst:


stianst:
Currently EAP 7.2 docs are not published so tests will not work. Will need to figure out when we merge this.

stianst:
Ran tests with internally staged EAP 7.2 docs and only one link failed, which is fixed in this PR

matthewhelmke (dismissed):
It is nice that only one link broke with the change to 7.2.
This LGTM.

matthewhelmke (approved):

PR #542, opened 2018-12-07
Fixed some links in the securing_apps chapter

entfrickler:
I think this should also been back merged into older versions of the documentation.

I tested the links with version 4.6.0.Final

### DIFF
--- a/securing_apps/topics/client-registration.adoc
+++ b/securing_apps/topics/client-registration.adoc
@@ -6,7 +6,7 @@ An admin can do this through the admin console (or admin REST endpoints), but cl
 registration service.
 
 The Client Registration Service provides built-in support for {project_name} Client Representations, OpenID Connect Client Meta Data and SAML Entity Descriptors.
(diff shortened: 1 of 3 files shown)


matthewhelmke (approved):
Thank you! We appreciate your submission.

Our process for docs PRs always includes two reviews: one from me from a writing standpoint (you pass easily as this has no writing changes for me to review) and one from engineering to check any changes or additions to technical details. They will check the URL changes and comment here.

stianst:
By default the updated paths are correct, but throughout the documentation we use paths relative to the default context-path for Keycloak which is /auth so we have intentionally left /auth out. Not sure if that was a good idea or not though.

matthewhelmke:
@stianst That clears the question up for me. I would prefer to put in the default context-path rather than leave it out, mainly because anyone who makes a change from the default should also be able to realize their path will differ, but those who have not made the change may stumble over things because they don't realize they need to add it...as in this PR.

stianst (approved):

PR #543, opened 2018-12-11
Add -c parameter

dometec:
Add -c parameter on jboss-cli invocation for run cli script while wildfly is running and some path fix.


matthewhelmke (approved):
Thank you for your contribution! We generally have two reviewers on docs submissions. I review based on writing standards, which don't really apply here since you are making code changes only (so you pass my review). We have a second review from engineering from a technical standpoint.

abstractj (approved):

PR #549, opened 2018-12-18
Release Notes need Known Issues

matthewhelmke:
I'm including a small number of formatting improvements and phrasing edits here, but really this PR is intended to include the Known Issues, as soon as I have them.

PR #550, opened 2018-12-18
Minor tweaks

jdoyleoss:
Tweaks to the order of standards and the tech preview blurb.


matthewhelmke:
I have taken the changes from here and added them into a commit in #549. I am closing this one (even though it is appreciated).

PR #551, opened 2018-12-24
KEYCLOAK-9172 Documentation for Identity Provider vk.com

wlami:
This PR adds documentation for setting up vk.com as Identity Provider. The corresponding feature is implemented in https://github.com/keycloak/keycloak/pull/5829.

There is one thing I am not happy about: in `server_admin/topics/identity-broker/social/vkontakte.adoc` line 15 I had to escape the registration URL using a backslash. Otherwise, the mvn build would throw an error during the test phase. This is because the URL is protected and one has to be signed in to access it. If you have any suggestion on how to make this link clickable, I would be very thankful.


matthewhelmke (approved):
Thank you for this submission. As you may have surmised, everyone has been out for the end of the year. That is the reason for my delayed response.

We have two reviews for docs submissions: one from me from a docs perspective and one from engineering. What you have submitted parallels the documentation for other social providers and looks good to me. Thank you! I'm sure someone from engineering will review this around the same time as they review your code submission.

wlami:
@matthewhelmke Thank you for the feedback. Do you have any idea on how to make the link in `vkontakte.adoc line 15` clickable, without breaking the maven build of the documentation?

matthewhelmke:
@wlami There is a filter to ignore some URLs, but I did not set it up. Let's see what @stian thinks when he is back from holiday.

PR #552, opened 2018-12-31
KEYCLOAK-9180: Added dependency information as as prerequisite.

lennartj:
Added trivial description of required POM dependencies to start working on the example.

### DIFF
--- a/server_development/topics/auth-spi.adoc
+++ b/server_development/topics/auth-spi.adoc
@@ -152,31 +152,92 @@ org.keycloak.examples.authenticator.AnotherProviderFactory
 
 This services/ file is used by Keycloak to scan the providers it has to load into the system. 
 
-To deploy this jar, just copy it to the providers directory. 
+To deploy this jar, just copy it to the providers directory.
+
+===== Implementing an Authenticator: Project Prerequisites
+
+Before implementing a custom Authenticator, we need to import the Keycloak dependency containing the Authenticator
+interface which is defined within the artifact `keycloak-server-spi-private` using the Keycloak release version.
+Also, the dependencies `keycloak-server-spi-private` and `keycloak-server-spi-private` are required to access other
+parts of the Keycloak server and inbound HTTP request parts such as Cookies.
+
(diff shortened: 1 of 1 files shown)


matthewhelmke (dismissed):
Wow, this is quite nice, thank you.

We have two reviews for docs contributions, one from me from a writing standpoint and another from engineering to confirm the details. Your contribution has passed mine.

server_development/topics/auth-spi.adoc
```diff
@@ -152,78 +152,204 @@ org.keycloak.examples.authenticator.AnotherProviderFactory
...
 
 [source,java]
 ----
+    /**
+     * Called if this Authenticator is <em>not</em> configured for the provided user (i.e. if the
+     * {@link #configuredFor(KeycloakSession, RealmModel, UserModel)} method returned <code>false</code>).
+     * This method defines required actions which must be performed by the user before this {@link Authenticator}
+     * can be used.
+     *
+     * {@inheritDoc}
```
lennartj: The build seems to treat the inheritDoc as a variable rather than as part of the javadoc.
Simply remove this line and we should be OK to merge - this is what the build log complains about.

```suggestion 
```

stianst:
Thanks for the PR. It's quite a nice update, but I'm not convinced about adding as it may be better to refer to quickstart for specifics. Documentation like this has a tendency to get outdated. At the very least versions in the docs would need to be templated not harcoded to an old version as it is now.

abstractj:
> Thanks for the PR. It's quite a nice update, but I'm not convinced about adding as it may be better to refer to quickstart for specifics. Documentation like this has a tendency to get outdated. At the very least versions in the docs would need to be templated not harcoded to an old version as it is now.

For more that I appreciate the contribution, as you said it may become outdated. @lennartj  would you like to contribute to the quickstarts?

abstractj:
It has been a while since we requested changes to this PR. While the submission is appreciated, we are not able to merge it without these changes. Because it has been so long, we are going to close the issue. Feel free to reopen with the changes included, if your time permits.

PR #569, opened 2019-01-21
KEYCLOAK-4989 Note about LDAP password hashing

mposolda:


maunzCache:
I'd like to disagree with that PR.
https://tools.ietf.org/html/rfc4519#section-2.41

From my understanding password are stored in LDAP "as given" (string octet). Which means LDAP does no hashing and Keycloak needs to generate the hash before storing a password there.

matthewhelmke (approved):
This is clearly worded and looks good to me.

mposolda:
@matthewhelmke Thanks, I think that's my 1st docs PR, which you approved for the 1st time :)

@getjack See my comment in the JIRA regarding MSAD, RHDS and FreeIPA: https://issues.jboss.org/browse/KEYCLOAK-4989?focusedCommentId=13418458&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13418458 . Also see the additional comment in the JIRA (right after my comment) from Konstantin Gribov regarding OpenLDAP. 

If you still disagree, please point me with some more details and I will try to update the PR accordingly.

quanah:
Actually, the problem appears to be that JBOSS is not correctly using an LDAP v3 modify operation, which then allows the server to hash the password per the server configuration.  A direct ldapmodify against the userPassword attribute bypasses the server's built-in hashing mechanisms.

See https://tools.ietf.org/html/rfc3062

quanah:
This commit should definitely be reverted as it is entirely erroneous.  The correct text would be something like:

``
JBOSS currently does not implement the LDAPv3 password modification operation as described in RFC3062, and instead incorrectly directly updates the userPassword attribute.  As a result, RFC compliant LDAP servers will not hash the password and will instead store it in cleartext form.  Until JBOSS is fixed to use the correct LDAPv3 methodology for password changes, it is advised to check the documentation for an RFC compliant LDAP server on how to configure it so that it will intercept these password change requests and hash the password.  For example: With OpenLDAP this can be achieved via the slapo-ppolicy module and the ppolicy_hash_cleartext directive.
``

PR #573, opened 2019-01-22
Fix theme archive extension from zip to jar

ghost:
Hi,

there is a tiny issue in the documentation, which did cost me some time to find - auto deployment of themes only works with JAR files, not with ZIP.


matthewhelmke (approved):
This looks good to me. Thank you. Someone from engineering will also take a look to double check.

stianst (approved):

stianst:
Thanks

PR #600, opened 2019-03-09
KEYCLOAK-6795 Silent refresh of tokens for implicit flow

dasniko:
Documentation for KEYCLOAK-6795

### DIFF
--- a/securing_apps/topics/oidc/javascript-adapter.adoc
+++ b/securing_apps/topics/oidc/javascript-adapter.adoc
@@ -144,6 +144,50 @@ keycloak.init({ flow: 'implicit' })
 One thing to note is that only an access token is provided and there is no refresh token. This means that once the access token has expired the application
 has to do the redirect to the {project_name} again to obtain a new access token.
 
+For SPAs, there's a the possibility to use a _silent refresh_ of the access token when using the implicit flow.
+This is achieved with the help of a hidden iframe in the application.
+Therefore, the init parameter `silentRefreshRedirectUri` needs to be provided on {project_name} initialization.
+This URI needs to be a valid endpoint in the application (and of course it must be configured as a valid redirect for the client in the {project_name} Administration Console):
+
+[source,javascript]
+----
+keycloak.init({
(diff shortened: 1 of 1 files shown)


marcelnem:
Hi, thank you for implementing this! I am looking forward to try this. I am not sure I completely understand the page in the iframe.

```
The page at the silent refresh uri is loaded in the iframe after successfully obtaining new tokens from the {project_name} server.
It has no other task than sending the received tokens to the main application and should only look like this:

[source,html]
----
<html><body>
  <script>
    parent.postMessage(location.href, location.origin)
  </script>
</body>
</html>
----
```

- Do we need to do something so that the page in the iframe gets the tokens from keycloak? If I just put the specified template page there, this page has no reason to get keyclok tokens, right? So I am guessing the page in the iframe must be protected by keycloak login too. But the Keycloak login will happen without forcing auser to write username/password because the session is still valid.



What mechanism is used to pass the token from page in the iframe to the parent page? 

>     parent.postMessage(location.href, location.origin)

I am guessing that the token is inside the URL of the page loaded in the iframe and the script of the parent page will extract it?

dasniko:
The `silent-refresh.html` page is a redirect page, the Keycloak server is redirecting to, after a successful login request. This login request is done in the iframe with a `prompt=none` and the redirectUrl of this silent-refresh.html. The page just passes the URL to the main window/script (keycloak.js) and it will be processed there. I've extended the `updateToken()` method of `keycloak.js` to achieve this, see https://github.com/keycloak/keycloak/pull/5932
The only thing you have to do in your app is to initialize the Javascript adapter with the `silentRefreshRedirectUrl` and to put the template into your app. It doesn't need to be secured, but it must be configured as a valid redirectUrl in Keycloak client configuration.

marcelnem:
Ok. I understand now. Thank you very much!

dasniko:
Failing build is due to other errors than my PR!

stianst:
Closing this as PR for code update and issue has been rejected.

PR #601, opened 2019-03-11
KEYCLOAK-6056 Map user by Subject Alternative Name (otherName) when a…

mposolda:
…uthenticating user with X509


mposolda:
Implementation PR: https://github.com/keycloak/keycloak/pull/5937

hmlnarik (approved):

PR #622, opened 2019-03-24
Update initial-user.adoc

lpetit-yseop:


stianst:
I prefer permitted over allowed, @andymunro wdyt?

andymunro (commented):
I agree with Stian.  Permitted is better than allow.  Permitted fits in with the concept of permissions.

Just on a short break from class.  I have to head back now.

lpetit-yseop:
OK then. As a non-native english speaker, I thought permitted was an old english term used inappropriately, but then I'm happy to be proven wrong. Closing the pull request.

PR #626, opened 2019-04-01
Fix link from Event Listener SPI to Service Provider Interfaces

daniel-shuy:
Link in <https://www.keycloak.org/docs/latest/server_development/index.html#_events> to <https://www.keycloak.org/docs/latest/server_development/index.html#_providers> is broken


stianst:
Link on https://www.keycloak.org/docs/latest/server_development/index.html#_events works for me, but link in docs generated from this PR (https://keycloak-docs.github.io/deploy-docs-pullrequest/PR/626/server_development/index.html) doesn't.

daniel-shuy:
Huh, that's odd, its working now. Never mind.

PR #627, opened 2019-04-01
KEYCLOAK-9966: Explain better how URL/Binding is selected for IDP iniated login

rmartinc:
Document a bit better how the IDP initiated logins selects final SP SAML processing URL and the binding (now it's possible to do that after KEYCLOAK-9068).
I'm very bad writing documentation, so any improvement you see please let me know.

### DIFF
--- a/server_admin/topics/clients/saml/idp-initiated-login.adoc
+++ b/server_admin/topics/clients/saml/idp-initiated-login.adoc
@@ -6,6 +6,15 @@ In the `Settings` tab for your client, you need to specify the `IDP Initiated SS
 This is a simple string with no whitespace in it.
 After this you can reference your client at the following URL: `root/auth/realms/{realm}/protocol/saml/clients/{url-name}`
 
+The IDP initiated login implementation prefers _POST_ over _REDIRECT_ binding (check <<saml-bindings, saml bindings>> for more information).
+Therefore the final binding and SP URL are selected in the following way:
+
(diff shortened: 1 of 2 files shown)


server_admin/topics/clients/saml/idp-initiated-login.adoc
```diff
@@ -6,6 +6,15 @@ In the `Settings` tab for your client, you need to specify the `IDP Initiated SS
 This is a simple string with no whitespace in it.
 After this you can reference your client at the following URL: `root/auth/realms/{realm}/protocol/saml/clients/{url-name}`
 
+The IDP initiated login prefers POST binding over REDIRECT (if using REDIRECT the SAML response is part of the URL, it is as a paremeter,
```
hmlnarik: The part in brackets is general recommendation, not specific to IdP-initiated login. Could you please move it to https://github.com/keycloak/keycloak-documentation/blob/master/server_admin/topics/sso-protocols/saml.adoc#post-binding?

hmlnarik (commented):
@rmartinc Thank you! Looks almost good, could you please only address a single inline comment? Then it would be ready to go.

rmartinc:
@hmlnarik I have moved that line to the other section, but, as that one is more general, I have added the two reasons why POST is usually preferred over REDIRECT (security and size). Regards!

hmlnarik (approved):
@rmartinc Thanks for the update, looks good to me now!

PR #632, opened 2019-04-08
fix-broken-heading

translucens:
The header of `mod_auth_openidc` is missing and its sentences are shown in Keycloak Gatekeeper section. So I fix its heading.


abstractj:
@translucens thanks for submitting this. But I'm afraid to say that changing the heading won't fix the issue, please run `mvn clean install` and take a look at the html file inside target folder.

You need to remove the asterisk from gatekeeper file:
```
==== Known Issues
* There is a known issue with the Keycloak server 4.7.0.Final in which Gatekeeper is unable to find the _client_id_ in the _aud_ claim. This is due to the fact the _client_id_ is not in the audience anymore. The workaround is to add the "Audience" protocol mapper to the client with the audience pointed to the _client_id_. For more information, see link:https://issues.jboss.org/browse/KEYCLOAK-8954[KEYCLOAK-8954].
```

translucens:
@abstractj thank you for your reviewing and kindly comment. I fixed as you advised, built documents on local environment and confirmed these documents are rendered as expected.

abstractj (approved):

PR #633, opened 2019-04-08
Remove reference to missing authenticator project example

cliffmaury:
examples/providers/authenticator is no more in repository as demo distribution has been removed (see KEYCLOAK-8437)


stianst (changes requested):
Example was deleted by mistake. We should either recover it or move it to quickstarts and update docs.

abstractj:
@cliffmaury thanks for submitting this. Although, like @stianst mentioned it, the example was removed by mistake. The PR to recover it is already there and hopefully will be included in further releases https://github.com/keycloak/keycloak/pull/6025.

For now I'm closing this.

PR #635, opened 2019-04-09
Update providers.adoc

tsabirgaliev:
Give some hints on how to setup a Maven project to get started

### DIFF
--- a/server_development/topics/providers.adoc
+++ b/server_development/topics/providers.adoc
@@ -5,6 +5,36 @@
 {project_name} is designed to cover most use-cases without requiring custom code, but we also want it to be customizable.
 To achieve this {project_name} has a number of Service Provider Interfaces (SPI) for which you can implement your own providers.
 
+=== Getting started
+
+{project_name} Service Provider Interfaces are available as Maven artifacts. Here is a `pom.xml` sample to get you started:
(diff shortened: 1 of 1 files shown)


tsabirgaliev:
I'm sorry, I can't get rid of empty line at the end, Github UI adds that :(

server_development/topics/providers.adoc
```diff
@@ -307,4 +337,4 @@ public class EjbExampleUserStorageProviderFactory
 
 === Available SPIs
 
-If you want to see list of all available SPIs at runtime, you can check `Server Info` page in admin console as described in <<_providers_admin_console,Admin Console>> section.
\ No newline at end of file
+If you want to see list of all available SPIs at runtime, you can check `Server Info` page in admin console as described in <<_providers_admin_console,Admin Console>> section.
```
andymunro: Please add "the" before 'list of all available' and 'Server Info' page

stianst (changes requested):
Not convinced this should be added to docs as it is covered better by QuickStarts. If we want to merge this version of BOM needs to be templated in using variable rather than hardcoded as it is now.

stianst:
Closing due to lack of feedback

PR #649, opened 2019-04-24
Fix mojibake

k-tamura:
`’`, `“` and `”` are used in the Keycloak guides. They cause mojibake.

For example:

- > database you???re using.
- > from the old server ???themes??? directory to the new server ???themes??? directory.

Please refer to the [Upgrading Guide](https://www.keycloak.org/docs/5.0/upgrading/index.html#manual-relational-database-migration).


k-tamura:
Closed this pull request because it is included in pull request #650.

PR #657, opened 2019-05-13
KEYCLOAK-10321 Connecting to an external database

slaskawi:
https://issues.jboss.org/browse/KEYCLOAK-10184

### DIFF
--- a/openshift/topics/advanced_concepts.adoc
+++ b/openshift/topics/advanced_concepts.adoc
@@ -281,6 +281,13 @@ If successful, the previous command will return the following output:
 route "sso" patched
 ----
 
+[[sso-connecting-to-an-external-database]]
+==== Connecting to an external database
+
+{project_name} can be configured to connect to an external (to OpenShift cluster) database. In order to achieve this, you need to modify the `sso-{database name}` Endpoints object to point to proper address. The procedure has been described in https://docs.openshift.com/container-platform/latest/dev_guide/expose_service/expose_internal_ip_service.html#manually-assign-ip-service[OpenShift manual].
+
+Tip: The easiest way to get started is to deploy {project_name} from a template and then modify the Endpoints object. You might also need to update some of the datasource configuration variables in the DeploymentConfig. Once you're done, just roll a new deployment out.
+
 [[sso-administrator-setup]]
 === Creating the Administrator Account for {project_name} Server
 


openshift/topics/advanced_concepts.adoc
```diff
@@ -281,6 +281,12 @@ If successful, the previous command will return the following output:
 route "sso" patched
 ----
 
+[[sso-connecting-to-an-external-database]]
```
iankko, line 284: @slaskawi This should be added to the `Tutorials` section of the book, rather than to the `Advanced Concepts` one.

openshift/topics/advanced_concepts.adoc
```diff
@@ -281,6 +281,12 @@ If successful, the previous command will return the following output:
...
+==== Connecting to an external database
+
+{project_name} can be configured to connect to an external (to OpenShift cluster) database. In order to achieve this, you need to modify the `sso-{database name}` Endpoint object to point to proper address. The procedure has been described in https://docs.openshift.com/container-platform/3.4/dev_guide/expose_service/expose_internal_ip_service.html#manually-assign-ip-service[OpenShift manual].
```
iankko: Please prefer using _`latest`_ version, when referring to OCP documentation, since that one will get automatically pointed to the next version, once it's available, e.g.:

```
https://docs.openshift.com/container-platform/latest/dev_guide/expose_service/expose_internal_ip_service.html#manually-assign-ip-service[OpenShift manual].
```

Besides that OCP v3.4 is an old version now.

iankko (changes requested):
Not sufficient detail, incomplete. Providing just link to official OCP documentation isn't sufficient.

Given the IP / hostname of the remote PostgreSQL server being known / provided, the change should provide complete scenario, how to get the RH-SSO pod running on OCP working with that. IOW set of 5-, 10-, 15- steps, that can be copy & pasted, and would work OOTB with minimal amount of changes. If the IP / hostname can be requested by the script, the whole setup / customized image would be produced just by calling / running that script. This can be done either utilizing the provided datasources example, or using the provided extensions post-configure example.

openshift/topics/advanced_concepts.adoc
hmlnarik: @iankko Why? Is this a tutorial?

slaskawi: As I explained in the [KEYCLOAK-10184](https://issues.jboss.org/browse/KEYCLOAK-10184), I would prefer to have a concise information about this topic. Unless you insist, I would like to keep it here (and not produce a longer tutorial).

openshift/topics/advanced_concepts.adoc
slaskawi: Fixed. Thanks!

openshift/topics/advanced_concepts.adoc
iankko: > @iankko Why? Is this a tutorial?

The `tutorials` subchapter would be a proper place in the case we would want to turn this into a tutorial (that was my original understanding). But since the real problem, why sharing an advanced link to official OCP documentation on how to configure external DB service isn't enough is [CLOUD-3198](https://issues.jboss.org/browse/CLOUD-3198), after the discussion / review, I am OK with providing a paragraph to the advanced section. The real issue is [CLOUD-3198](https://issues.jboss.org/browse/CLOUD-3198), and once that one is corrected, link to official documentation is enough.

iankko (approved):
The current version LGTM.

hmlnarik (approved):

PR #661, opened 2019-05-17
KEYCLOAK-10288 Fix warnings in build log, and one ccutils error

ASzc:
The warnings in the log will cause all automatic builds to be in warning status instead of success.

This PR is also to try out changes to dawbrn.

### DIFF
--- a/header-maven-plugin/pom.xml
+++ b/header-maven-plugin/pom.xml
@@ -50,7 +50,7 @@
                 <artifactId>maven-plugin-plugin</artifactId>
                 <version>3.4</version>
                 <configuration>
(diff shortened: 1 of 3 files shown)


abstractj:
@ASzc could you please provide the Jira id and also the steps to reproduce?

ASzc:
For the warnings, just run `mvn package` over everything. You'll see them disappear with the addition of the two POM changes.

The third change covers an actual error: a crossreference to a non-existing ref. The error is in the ccutils (RH-SSO) documentation build, so you'll need that internal tool to reproduce the error. I'll send you the link internally.

abstractj (approved):

PR #679, opened 2019-06-13
Add attribute to refill value in template

k-tamura:
Please refer to [the thread](https://lists.jboss.org/pipermail/keycloak-user/2015-August/002943.html).


k-tamura:
refill

stianst (approved):

PR #680, opened 2019-06-13
[KEYCLOAK-10443] - Define a global decision strategy for resource servers

pedroigor:


abstractj (approved):
I manage to run it locally and LGTM

PR #712, opened 2019-08-13
KEYCLOAK-6768 Signed and Encrypted ID Token Support

tnorimat:
This documentation is for
* PR :  https://github.com/keycloak/keycloak/pull/6237
* JIRA : https://issues.jboss.org/browse/KEYCLOAK-6768

### DIFF
--- a/server_admin/topics/clients/client-oidc.adoc
+++ b/server_admin/topics/clients/client-oidc.adoc
@@ -164,7 +164,35 @@ The administrator can select the following three options:
 
 Please see https://tools.ietf.org/html/rfc7636[RFC 7636 Proof Key for Code Exchange by OAuth Public Clients] for more details.
 
+[[_jwe-id-token-encryption]]
+*Signed and Encrypted ID Token Support*
 
+{project_name} can encrypt ID token according to https://tools.ietf.org/html/rfc7516[Json Web Encryption (JWE)] specification. The administrator can determine whether encrypting ID token or not per client. This feature is disabled as default.
+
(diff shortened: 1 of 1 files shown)


mposolda (approved):
@tnorimat Thanks, I am approving.

I have one small suggestion: IMO this feature deserves to be mentioned in the release notes. So if you can also add few notes to the release notes (file changes.adoc), it will be a bonus point :) I think some short notice (1 or 2 sentences would be fine, see existing release notes for more details). IMO this is not a blocker, so I am approving anyway.

I am adding @andymunro from our documentation team to review as well.

andymunro (approved):

tnorimat:
@mposolda Thank you the suggestion of adding notes on the release note.

I think it might be better to add this notes by some maintainers. I would be happy if doing so.

As for the contents of this notes, what do you think about the following?

### Signed and Encrypted ID Token Support

Keycloak can support the signed and encrypted ID token according to Json Web Encryption (JWE) specification. 

JWE specification defines several ways of realizing signed and encrypted JWT and keycloak only supports some of them. These are combinations of RSA1_5 or RSA-OAEP as Key Encryption algorithm and A128CBC-HS256 or A128GCM as Content Encryption algorithm.

To use this feature, the client application need to manage key pairs of public-key cryptography and provide public keys of them to keycloak according to Json Web Keys (JWK) specification.

mposolda:
@andymunro Thanks for the review

@tnorimat Thanks for the suggestion! I think I will use just the first sentence of it with some additional details found in the documentation.

PR #732, opened 2019-09-06
Example configuration for running as sidecar container in Kubernetes

transducer:


stianst:
Hi, thanks for your PR and sorry for the late reply. I'm not 100% convinced that it's ideal to provide so much detail in the documentation, would probably think some sort of QuickStart would be better. Maybe we could add a QuickStart around Gatekeeper on https://github.com/keycloak/keycloak-quickstarts including templates for Kube and OpenShift. @abstractj WDYT?

abstractj:
@stianst @transducer a quickstart would be fantastic +1 about it. Often code at the docs get outdated in few month and it's hard to test.

stianst (changes requested):
Documentation shouldn't contain large examples like this, rather we should have a quickstart for Gatekeeper around this.

stianst:
Thanks for the PR, but we are not able to maintain large examples like this within the documentation.

PR #733, opened 2019-09-06
KEYCLOAK-11072 Document Vault SPI

hmlnarik:
Documentation for keycloak/keycloak#6294


server_development/topics/vault.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+=== Vault provider
+
+You can use a vault SPI from `org.keycloak.vault` package to write custom extension for {project_name} to connect to arbitrary vault implementation.
+
+The built-in `plaintext` provider is an example of the implementation of this SPI. In general the following rules apply:
+
+* The vault provider should respect realm, i.e. an expression `${vault.secret-id}` would evaluate generally to different values of a secret `secret-id`, depending on whether it was used in a realm _A_ or realm _B_. This is to prevent secret leaking across realms. To differentiate between realms, the realm needs to be passed to the created `VaultProvider` instance from `VaultProviderFactory.create()` method where it is available from its the `KeycloakSession` parameter.
+
+* The vault provider needs to implement a single method `obtainSecret` that returns either `byte[]` or `ByteBuffer` filled with the secret contents for the given secret name. Note that this buffer would be discarded after usage as explained below.
```
sguilhen: It returns a VaultRawSecret that in turn can hand out the secret as a byte[] or enclosed in a ByteBuffer.

server_admin/topics/realms/email.adoc
```diff
@@ -36,5 +36,5 @@ As emails are used for recovering usernames and passwords it's recommended to us
 To enable SSL click on `Enable SSL` or to enable TLS click on `Enable TLS`.
 You will most likely also need to change the `Port` (the default port for SSL/TLS is 465).
 
-If your SMTP server requires authentication click on `Enable Authentication` and insert the `Username` and `Password`.
+If your SMTP server requires authentication click on `Enable Authentication` and insert the `Username` and `Password`. The value of `Password` field can refer a value from an external <<_vault-administration,vault>>.
```
sguilhen, line 40: I was wondering if a similar section existed for LDAP bind credential and the client secrets but I couldn't find any, so I'm guessing this is all we need for now.

sguilhen (commented):
Shouldn't we have a section in changes.adoc with a brief description of the new vault SPI being available?

server_development/topics/vault.adoc
hmlnarik: Yes, that will be a better formulation, thanks!

server_admin/topics/vault.adoc
```diff
@@ -0,0 +1,57 @@
+[[_vault-administration]]
+
+== Using Vault to Obtain Secrets
+
+Several fields in the administration support obtaining the value of a secret from an external vault.
+
+To obtain a secret from a vault instead of entering it directly, enter
+the following specially crafted string into the appropriate field:
+`**${vault.**_vault.name_**}**` where you replace the `_vault.name_`
```
mhajas:
```suggestion
`**${vault.**_vault.entry.name_**}**` where you replace the `_vault.entry.name_`
```

mhajas (commented):
Thank you for PR @hmlnarik. I added one minor comment in the code. Should we document also VaultTranscriber?

mhajas (dismissed):

martin-kanis (dismissed):

server_admin/topics/vault.adoc
```diff
@@ -0,0 +1,57 @@
...
+To use a vault, a vault provider must be registered within {project_name}.
+It is possible to either use a built-in provider described below or
+implement your own provider. See the link:{developerguide_link}[{developerguide_name}] for more information.
+
+=== Kubernetes / OpenShift plaintext vault provider
+
+{project_name} supports vault implementation for https://kubernetes.io/docs/concepts/configuration/secret/[Kubernetes secrets]. These secrets
+can be mounted as data volumes, and they appear as a directory with flat file structure, where each secret is represented by a file whose name is secret name, and contents of that file is the secret value.
+
+The files within this directory have to be named as secret name prefixed by realm name and an underscore. All underscores within the secret name or realm name have to be doubled in the file name. For example, for a field within a realm called `sso_realm`, a reference to a secret with name `secret-name` would be written as `${vault.secret-name}`, and the file name looked up would be `sso+++__+++realm+++_+++secret-name`.
```
stianst: sso+++__+++realm+++_+++secret-name - is sso__realm, but I assume it should be sso_realm

server_admin/topics/vault.adoc
```diff
@@ -0,0 +1,57 @@
+[[_vault-administration]]
+
+== Using Vault to Obtain Secrets
+
+Several fields in the administration support obtaining the value of a secret from an external vault.
+
+To obtain a secret from a vault instead of entering it directly, enter
+the following specially crafted string into the appropriate field:
+`**${vault.**_vault.entry.name_**}**` where you replace the `_vault.entry.name_`
```
stianst: maybe "vault.entry-name" would be better? Using dots in vault.entry.name makes it a bit confusing as you may think it's a hierarchy of sorts

server_admin/topics/vault.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+Here is the equivalent configuration using CLI commands:
+
+[source,bash]
+----
+/subsystem=keycloak-server/spi=vault/:add
+/subsystem=keycloak-server/spi=vault/provider=plaintext/:add(enabled=true,properties={dir => "${jboss.home.dir}/standalone/configuration/vault"})
+----
+
+NOTE: There is at most one vault provider active per {project_name} instance at any given time.
```
stianst: Not per-instance, but rather per cluster.

server_development/topics/vault.adoc
```diff
@@ -0,0 +1,44 @@
+[[_vault-spi]]
+== Vault SPI
+
+=== Vault provider
+
+You can use a vault SPI from `org.keycloak.vault` package to write custom extension for {project_name} to connect to arbitrary vault implementation.
+
+The built-in `plaintext` provider is an example of the implementation of this SPI. In general the following rules apply:
+
+* The vault provider should respect realm, i.e. an expression `${vault.secret-id}` would evaluate generally to different values of a secret `secret-id`, depending on whether it was used in a realm _A_ or realm _B_. This is to prevent secret leaking across realms. To differentiate between realms, the realm needs to be passed to the created `VaultProvider` instance from `VaultProviderFactory.create()` method where it is available from its the `KeycloakSession` parameter.
```
stianst: I would rewrite this to something along the lines of "If you want to isolate or limit what secrets can be retrieved by a realm your provider should take into account the realm name when looking up secrets, for example by prefixing entries with the realm name.

server_admin/topics/vault.adoc
hmlnarik: Here it really needs to be escaped in the file name so that it is distignuishable where the realm name ends.

server_admin/topics/vault.adoc
hmlnarik: +1

server_admin/topics/vault.adoc
hmlnarik: That's a good point to prevent confusion. It is per instance (the provider) but within the cluster the vault definition needs to be consistent. Vault data are not shared across the cluster via infinispan to prevent leaking, it needs to be configured on all instances separately.

server_development/topics/vault.adoc
hmlnarik: I like this one better, thanks!

server_admin/topics/vault.adoc
stianst: Hmm... after re-reading the paragraph several times I finally understand it, but find it confusing and error prone

stianst: A different delimiter between realm name and secret name would be better

stianst: What about secrets/realm-name/secret-name?? Then it's a directory per-realm. Could also have secrets at the top-level that is shared between realms.

server_admin/topics/vault.adoc
stianst: Now I'm confused. Are you talking about provider instances, not server instances?

stianst:
One more think I'm not hugely keen on is the name plaintext for the Kubernetes provider. Should be files or something.

server_admin/topics/vault.adoc
hmlnarik: Sadly this is not user-friendly in K8s, since every entry within a secret would have to have the `path` mapping set for this to work, and I am not certain that OpenShift even supports that from UI or can only set that from within a yaml

server_admin/topics/vault.adoc
hmlnarik: Yes, provider instance within a server. Server instances needs to be configured consistently inside cluster

stianst (dismissed):
Approving, but would like a discussion around plaintext provider.name and the path syntax for.mounting June secrets

sguilhen (dismissed):
Looks good to me!

server_admin/topics/realms/email.adoc
```diff
@@ -36,5 +36,5 @@ As emails are used for recovering usernames and passwords it's recommended to us
...
```
andymunro: Suggest value of THE "Password" field..

server_admin/topics/vault.adoc
```diff
@@ -0,0 +1,59 @@
...
+== Using Vault to Obtain Secrets
+
+Several fields in the administration support obtaining the value of a secret from an external vault.
+
+To obtain a secret from a vault instead of entering it directly, enter
+the following specially crafted string into the appropriate field:
+`**${vault.**_entry-name_**}**` where you replace the `_entry-name_`
+with the name of the secret as recognized by the vault.
+
+Currently, secret can be obtained from vault in the following fields:
```
andymunro: Suggest from THE vault...

server_admin/topics/vault.adoc
```diff
@@ -0,0 +1,59 @@
...
+  In _Client Secret_ inside identity provider <<_identity_broker_oidc,OpenID Connect Config>>
+
+To use a vault, a vault provider must be registered within {project_name}.
+It is possible to either use a built-in provider described below or
+implement your own provider. See the link:{developerguide_link}[{developerguide_name}] for more information.
+
+=== Kubernetes / OpenShift plaintext vault provider
+
+{project_name} supports vault implementation for https://kubernetes.io/docs/concepts/configuration/secret/[Kubernetes secrets]. These secrets
+can be mounted as data volumes, and they appear as a directory with flat file structure, where each secret is represented by a file whose name is secret name, and contents of that file is the secret value.
```
andymunro: with A flat file structure

server_admin/topics/vault.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+The files within this directory have to be named as secret name prefixed by realm name and an underscore. All underscores within the secret name or realm name have to be doubled in the file name. For example, for a field within a realm called `sso_realm`, a reference to a secret with name `secret-name` would be written as `${vault.secret-name}`, and the file name looked up would be `sso+++__+++realm+++_+++secret-name` (note the underscore doubled in realm name).
```
andymunro: suggest THE secret name.... THE realm name...

server_admin/topics/vault.adoc
andymunro: Currently, THE secret...

server_development/topics/vault.adoc
```diff
@@ -0,0 +1,49 @@
+[[_vault-spi]]
+== Vault SPI
+
+=== Vault provider
+
+You can use a vault SPI from `org.keycloak.vault` package to write custom extension for {project_name} to connect to arbitrary vault implementation.
+
+The built-in `plaintext` provider is an example of the implementation of this SPI. In general the following rules apply:
+
+* To prevent secret leaking across realms, you mey want to isolate or limit what secrets can be retrieved by a realm.
```
andymunro: to prevent A secret FROM leaking...

server_development/topics/vault.adoc
```diff
@@ -0,0 +1,49 @@
...
+The example uses `KeycloakSession.vault()` as the entrypoint for accessing
+the secrets. Using the `VaultProvider.obtainSecret` method directly is indeed
+also possible. However the `vault()` method has the benefit of ability
+to interpret the raw secret (which is generally a byte array)
+as a character array (via `vault().getCharSecret()`) or a `String`
+(via `vault().getStringSecret()`) in addition to obtaining the original
+uninterpreted value (via `vault().getRawSecret()` method).
+
+Note that since `String` objects are immutable, their content cannot be discarded
+by overriding with random garbage. Even though measures have been taken in default
```
andymunro: THE default...

server_admin/topics/vault.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+To use this type of secret store, you have to declare the `plaintext` vault provider in standalone.xml, and set its parameter for directory which contains the mounted volume. The following example shows the `plaintext`
```
andymunro: parameter for THE directory THAT contains...

mposolda (approved):

PR #736, opened 2019-09-11
KEYCLOAK-4722 Support reCAPTCHA v3

codaabraham:
- Updated document for reCAPTCHA v3 support


spetar:
Hello, in which Keycloak version is this issue published ?

PR #742, opened 2019-09-17
 [KEYCLOAK-11436] Update JavaScript adapter documentation to prefer native promises

jonkoops:


jonkoops:
Since native Promises are not supported by all platforms (such as Internet Explorer). It would make sense to include a section where the user is informed of which polyfills may have to be included.

However I cannot find any information about which browsers are supported by Keycloak in the documentation.

andymunro (dismissed):

jonkoops:
No idea why the external link check is failing, seems to be a fluke.

securing_apps/topics/oidc/javascript-adapter.adoc
```diff
@@ -254,6 +270,14 @@ Example polyfill libraries:
 * HTML5 History - https://github.com/devote/HTML5-History-API
 * Promise - https://github.com/stefanpenner/es6-promise
 
+==== A note about `promiseType`
+
+Throughout the examples in the documentation you will find `promiseType: 'native'` specified wherever Keycloak is initialized.
```
andymunro: Should Keycloak be {project_name} throughout this content?
Otherwise, looks good.

jonkoops: Good catch, I've added it in the places where it seemed appropriate. It also seems like there are a lot of instances where this is not consistently adhered to as well, but that is outside of the scope of this PR.

andymunro (dismissed):
Thanks. Looks good.

jonkoops:
@stianst should we go ahead and merge this into the documentation?

jonkoops:
I've updated the code with an example of how to get the right API for native promises together with TypeScript. I've also replaced all instances where the function constructor of Keycloak was used with a preference towards using `new Keycloak()` to make the documentation more consistent.

jonkoops:
Looks like `[source,typescript]` is not supported, changed it back to `[source,javascript]`.

jonkoops:
@andymunro Could I ask you for one more review?

stianst (approved):

PR #753, opened 2019-10-15
Amendment for common Tomcat/SAML pitfall

dheuvels:
As already proposed in https://lists.jboss.org/pipermail/keycloak-user/2018-May/013984.html


hmlnarik (approved):
Thanks @dheuvels!

PR #772, opened 2019-11-07
Document proxy authentication support

k-tamura:
Document the proxy authentication support for proxy-mappings introduced by KEYCLOAK-10659.

See also:
https://github.com/keycloak/keycloak/pull/6111

### DIFF
--- a/server_installation/topics/network/outgoing.adoc
+++ b/server_installation/topics/network/outgoing.adoc
@@ -74,6 +74,14 @@ To determine the proxy for an outgoing HTTP request the target hostname is match
 hostname patterns. The first matching pattern determines the proxy-uri to use.
 If none of the configured patterns match for the given hostname then no proxy is used.
 
+If the proxy server requires authentication, include the proxy user's credentials in this format `username:password@`. 
+For example:
+
+[source]
+----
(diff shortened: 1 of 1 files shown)


server_installation/topics/network/outgoing.adoc
```diff
@@ -74,6 +74,13 @@ To determine the proxy for an outgoing HTTP request the target hostname is match
 hostname patterns. The first matching pattern determines the proxy-uri to use.
 If none of the configured patterns match for the given hostname then no proxy is used.
 
+If the proxy server requires authentication, include the proxy user's credentials like `username:password@`, e.g.:
```
andymunro: Instead of "like" please use "in this format"
Also, always replace "e.g."  with For example.  And please make it a new sentence.

andymunro (changes requested):
I added single comments.

server_installation/topics/network/outgoing.adoc
k-tamura: OK. I understand.

k-tamura:
Updated and rebased as @andymunro pointed out:

```
-If the proxy server requires authentication, include the proxy user's credentials like `username:password@`, e.g.:
+If the proxy server requires authentication, include the proxy user's credentials in this format `username:password@`.
+For example:
```

hmlnarik (approved):

PR #774, opened 2019-11-07
Node proxy docs updates

abstractj:


tests/src/test/resources/ignored-links
```diff
@@ -27,4 +27,5 @@ http://127.0.0.1:3000/oauth/callback
 https://api.linkedin.com/v2/me
 https://raw.githubusercontent.com/jboss-container-images/redhat-sso-7-openshift-image/sso-cd-dev/templates/${resource}
 https://raw.githubusercontent.com/jboss-container-images/redhat-sso-7-openshift-image/sso73-dev/templates/${resource}
-https://issues.jboss.org/secure/CreateIssueDetails*
\ No newline at end of file
+https://issues.jboss.org/secure/CreateIssueDetails*
+https://expressjs.com/en/guide/behind-proxies.html
```
stianst: Why are you ignoring the link? This means we won't know if it breaks.

abstractj: @stianst it is breaking at the moment, even though, the link works it returns HTTP 403 on tests. So I assumed that was our way to solve this.

andymunro (approved):

andymunro (approved):

tests/src/test/resources/ignored-links
stianst: No, that's just ignoring the issue ;)

The ignored-links are links that we simply can't test, such as example URLs and URLs that require authentication.

Not sure why it would return a 403, perhaps we're missing some headers or something in the request?

abstractj:
@stianst are we good to merge after the rebase?

stianst:
@abstractj we need to make sure the URL can be tested as otherwise we don't know if it stops working.

andymunro (dismissed):

abstractj:
@stianst I can remove that from ignore links, but tests will fail even though the URL exists.

PR #781, opened 2019-11-13
Admin REST API with client_secret

danifr:
Authenticating against Admin REST API with `client_id` and `client_secret`

### DIFF
--- a/server_development/topics/admin-rest-api.adoc
+++ b/server_development/topics/admin-rest-api.adoc
@@ -10,7 +10,9 @@ A token can be obtained by enabling authenticating to your application with {pro
 
 For complete documentation see {apidocs_link}[{apidocs_name}].
 
-=== Example using CURL
+=== Examples using CURL
+
+==== Authenticate with username and password
 
 Obtain access token for user in the realm `master` with username `admin` and password `password`:
 [source,bash]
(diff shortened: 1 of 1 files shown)


server_development/topics/admin-rest-api.adoc
```diff
@@ -10,7 +10,32 @@ A token can be obtained by enabling authenticating to your application with {pro
 
 For complete documentation see {apidocs_link}[{apidocs_name}].
 
-=== Example using CURL
+=== Examples using CURL
+
+==== Authenticating with `client_id` and `client_secret`
```
stianst: I'd change this to "Authenticate with a service account" and move it below the "Authenticate with username and password" example

danifr:
Shall we merge this? There are a few questions on StackOverflow about this topic and it would be nice if there is a solution from the official docs

server_development/topics/admin-rest-api.adoc
```diff
@@ -37,6 +39,29 @@ curl \
...
 ----
 
+==== Authenticate with a service account
+
+Before being able to authenticate against the Admin REST API using a `client_id` and a `client_secret` you need to make sure the client is configured as it follows:
+
+* `client_id` is a **confidential** client that belongs to the realm *master*
+* `client_id` has `Service Accounts Enabled` option enabled
+* `client_id` has a custom "Hardcoded claim" mapper
+  ** Token Claim Name: `azp`
```
stianst: Setting the azp here is not valid, and is misleading. The client should have scope on the required roles within security-admin-console, limited to what it needs. Then it should have the 'audience resolve' mapper added so the token has an audience on security-admin-console. which will be added automatically as it as a scope on the roles for that client.

stianst (changes requested):

server_development/topics/admin-rest-api.adoc
danifr: Hi Stian,

thanks for your comment. I will fix it.

>  The client should have scope on the required roles within security-admin-console, limited to what it needs

Can you clarify what you mean by this? I am not sure how I can translate that sentence into keycloak config (I am trying to reproduce what you say so I can eventually document it in plain english)

stianst: The steps should create a new client (not re-use admin-client client), remove full scope allowed, and select what roles it needs from the security-admin-console, then it should have the audience resolve mapper added. That will mean it has access to the roles it needs + the correct audience.

danifr: Thanks for your comments. I just updated my PR.

danifr:
Build is not passing. This commit seems to fix it though https://github.com/keycloak/keycloak-documentation/pull/918/commits/a48c7f10a2c199f5bac716059d306369d2a53a01

I will rebase from master after it gets merged.

stianst (approved):

PR #783, opened 2019-11-15
KEYCLOAK-7010 Configure JAVA_OPTS onOpenShift

andymunro:
Changing the link to remove the project_doc_base_url variable which caused problems. The new link will work only in the 7.3.x documentation since we are not issuing CD releases.  That link does include a project_version variable, so it will work at 7.4.  I hope this does not break anything.

Also, changing command to show sso-mysql, not sso73-mysql.
This PR replaces a previous one that I had intended to update: https://github.com/keycloak/keycloak-documentation/pull/759


andymunro:
I switched branches accidentally, which created a new PR, However, I have reapplied the changes to the existing PR: https://github.com/keycloak/keycloak-documentation/pull/759

PR #817, opened 2020-01-09
KEYCLOAK-12174 WebAuthn passwordless support

mposolda:


mposolda:
Code PR: https://github.com/keycloak/keycloak/pull/6649

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
-===== Edit Registered WebAuthn Authenticator
+The full configuration of the flow will look like this:
 
-A user can edit the following information :
+image:images/webauthn-passwordless-flow.png[]
 
-- Label (WebAuthn authenticator's label the user entered on registering it)
+You can now add `WebAuthn Register Passwordless` as required action to some user to test this. During first authentication, user will
+be still required to use password and his second-factor WebAuthn credential. However once user registers the Passwordless WebAuthn
+credential, he will be able to choose during future authentications. If he use his WebAuthn Passwordless credential, he won't need to
```
tnorimat: `If he use his WebAuthn Passwordless credential,` might be `If he uses his WebAuthn Passwordless credential,`

tnorimat (commented):
I've added one comment. Could you check it?

server_admin/topics/authentication/webauthn.adoc
mposolda: @tnorimat Fixed, Thanks!

mposolda:
@tnorimat Done

tnorimat (dismissed):
LGTM

mposolda:
@tnorimat Thanks

@andymunro Do you have a chance to review this PR please?

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -12,22 +12,17 @@ NOTE: Whether WebAuthn's operations succeed depends on a user's WebAuthn support
...
+[[_webauthn-register]]
 ===== Enable Webauthn Authenticator Registration
 
 An administrator carries out the following operations on the `Admin Console` :
 
 - Open the `Authentication -> Required Actions` tab.
 - Click `Register`.
 - Select `Webauthn Register` as `Required Action`.
-- Mark `Enabled` and `Default Action` checkbox.
+- Mark `Enabled` checkbox. Optionally mark `Default Action` checkbox if you want all newly created users to be required to register WebAuthn credential.
```
andymunro: change "newly" to "new"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -157,7 +155,8 @@ The appropriate method to register a WebAuthn authenticator depends on if the us
 
 New user::
 
-A new user carries out the following operations :
+In case that `WebAuthn Register` required action is set as `Default Action` in a realm, a new users will be always required to
```
andymunro: If the `WebAuthn Register` required action is set as `Default Action` in a realm, new users are required to
set up the WebAuthn security key after the first successful login. A new user carries out the following operations :

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
 
 Existing user::
 
-When existing users try to log in, they are required to register their WebAuthn authenticator automatically :
+In case that `WebAuthn Authenticator` is setup as required as shown in the first example, then when existing users try to log in,
```
andymunro: If `WebAuthn Authenticator` is set up as required as shown in the first example, then when existing users try to log in,

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
 - Open the login form.
 - Fill in items, click `Save` and  click `Login`.
 - When the users log in, they are required to register their WebAuthn authenticator.
 - After successful registration, the user's browser asks the user to enter the text as their just registered WebAuthn authenticator's label.
 
-===== View Registered WebAuthn Authenticator
+==== Passwordless WebAuthn
+
+WebAuthn is often used for two-factor authentication, however it can be desired to use it also as first factor authentication. In this case,
+user with `passwordless` WebAuthn credential will be able to authenticate to {project_name} without password. {project_name} allows to use WebAuthn
```
andymunro: Add "a" before "user" and "password"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+as both passwordless and two-factor authentication mechanism in the context of single realm and even in the context of single authentication flow.
```
andymunro: as both the passwordless and two-factor authentication mechanism in the context of a single realm and even in the context of a single authentication flow.

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+
+Administrator may typically require that Security Keys registered by users for the WebAuthn passwordless authentication must meet different
+(usually stronger) requirements. For example those security keys may require users to authenticate to that security key using PIN, or the
```
andymunro: For example, <-- add a comma

add "a" before PIN

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+security key should be attested with stronger certificate authority.
+
+Because of this, {project_name} allows administrator to configure separate `WebAuthn Passwordless Policy`. There is also separate required action
```
andymunro: Because of this situation,  <-- always follow "this" with a noun.

andymunro: add "a" before "also"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+of type `Webauthn Register Passwordless` and separate authenticator of type `WebAuthn Passwordless Authenticator`.
+
+===== Setup
+
+The setup procedure of WebAuthn passwordless support is the following :
+
+* Register new required action for WebAuthn passwordless support. Use the same steps as described <<_webauthn-register, above>>
```
andymunro, line 195: add "for" after "as described"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+with the only difference, that you need to register action called `Webauthn Register Passwordless`.
```
andymunro: add "the" before "action"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+
+* Configure the policy. You can use same steps and configuration options as described <<_webauthn-policy, above>>, however you
```
andymunro, line 198: add "for" after "as described"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+need to configure them in the admin console in the tab `WebAuthn Passwordless Policy`. You can configure this policy as you want, however
+typically the requirements for security key will be stronger than for two-factor policy. For example `User Verification Requirement` can
```
andymunro: add "the" before  for "security key" and "two-factor policy" and 'User Verification Requirement'

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+be set to `Required` when configure passwordless policy.
```
andymunro: when you configure the passwordless policy.

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+
+* Finally configure the authentication flow. Let's assume that we will use same flow called `WebAuthn Browser` as described
+<<_webauthn-authenticator-setup, above>>, but
+we will configure it like this:
```
andymunro: "as follows" instead of "like this"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+
+** The `WebAuthn Browser Forms` subflow will contain `Username Form` as first authenticator. This means that user will provide just
```
andymunro: add "the" before "first"
This setting means that the user will provide just his or her username as the first step.

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+his username as first step.
+
+** There will be required subflow, which can be named for example `Passwordless Or Two-factor` . This indicates that user can
```
andymunro: add "a" before "required"
Add "setting" after "This"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+authenticate either with Passwordless WebAuthn credential or with Two-factor authentication.
 
-A user carries out the following operations on the <<_account-service, `User Account Service`>>  :
+** Flow will contain `WebAuthn Passwordless Authenticator` as first alternative.
 
-- View the `Account` page.
+** Second alternative will be subflow named for example `Password And Two-factor Webauthn`. This subflow will contain `Password Form` and
```
andymunro: The second alternative will be a subflow named...

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+`WebAuthn Authenticator`.
 
-===== Edit Registered WebAuthn Authenticator
+The full configuration of the flow will look like this:
 
-A user can edit the following information :
+image:images/webauthn-passwordless-flow.png[]
 
-- Label (WebAuthn authenticator's label the user entered on registering it)
+You can now add `WebAuthn Register Passwordless` as required action to some user to test this. During first authentication, user will
```
andymunro: add "the" before "required" and "first" and "user"

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+be still required to use password and his second-factor WebAuthn credential. However once user registers the Passwordless WebAuthn
```
andymunro: be still required to use the password and second-factor WebAuthn credential. However, once the user registers the credentials, that user will be able...

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+credential, he will be able to choose during future authentications. If he uses his WebAuthn Passwordless credential, he won't need to
```
andymunro: If he uses his or her WebAuthn Passwordless credential, he won't need to provide the password...

andymunro (changes requested):
I suggested a few changes inline, Marek.

server_admin/topics/authentication/webauthn.adoc
mposolda: Hmmm... I am not good at english, but sentence "There is a also separate required action.." sounds really strange?

Did you meant:
add "a" INSTEAD OF "also" ? 

So the sentence will be like "There is a separate required action ..." .

server_admin/topics/authentication/webauthn.adoc
mposolda: Hmm... I am not sure about this one too. The snippet "<<_webauthn-register, above>>" is a link, which  is resolved as a text with single word "above" . So the resolved text after docs compilation currently looks like: `Use the same steps as described above with the only difference, ...`

With what you proposed, it will look like `Use the same steps as described for above with the only difference, ...`

Is it really correct?

server_admin/topics/authentication/webauthn.adoc
mposolda: I have same concern as above for this one.

mposolda (commented):
@andymunro Thanks a lot! I think I addressed most of your comments. I've just replied to 3 comments and hence I did not address those in a way you proposed.

For the first, I used the sentence "There is a separate required action ..." as I guess that is what you really meant in your comment?

Regarding my second and third comment, I did not address anything. Let me know if I am not correct and if you really meant to use the sentence like "Use the same steps as described for above with the only difference, ..." 

Thanks!

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+we will configure it as follows:
+
+** The `WebAuthn Browser Forms` subflow will contain `Username Form` as the first authenticator. This setting means that the user
+will provide just his or her username as the first step.
+
+** There will be a required subflow, which can be named for example `Passwordless Or Two-factor` . This setting indicates that user can
+authenticate either with Passwordless WebAuthn credential or with Two-factor authentication.
 
-A user carries out the following operations on the <<_account-service, `User Account Service`>>  :
+** Flow will contain `WebAuthn Passwordless Authenticator` as first alternative.
```
iankko: Missing `the` here? _** Flow will contain `WebAuthn Passwordless Authenticator` as the first alternative._?

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
 
-- View the `Account` page.
+** The second alternative will be a subflow named for example `Password And Two-factor Webauthn`. This subflow will contain `Password Form` and
```
iankko: _will contain a `Password Form` and a `WebAuthn Authenticator`._?

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+`WebAuthn Authenticator`.
 
-===== Edit Registered WebAuthn Authenticator
+The full configuration of the flow will look like this:
```
iankko: Replace _full_ with _complete_ or with _final_?
Replace _will look like this:_ with _will look like the following:_?

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
 
-A user can edit the following information :
+image:images/webauthn-passwordless-flow.png[]
 
-- Label (WebAuthn authenticator's label the user entered on registering it)
+You can now add `WebAuthn Register Passwordless` as the required action to some user to test this. During the first authentication, the user will
```
iankko: Maybe replace _as the required action to some user_ with _as the required action to some user, already known to {project_name},_ to test this? (This assumes Keycloak allows to set the `WebAuthn Register Passwordless` RA to some of the users, existing in Keycloak / already known to Keycloak. This differs e.g. from the `WebAuthn Register` RA action, which can be assigned solely to newly registered users AFAICT). 

But if the above assumption is wrong (IOW the new `WebAuthn Register Passwordless` RA can be only configured for newly registered Keycloak users too), I would complete remove the `to some user to test this.` part (IOW the sentence to be _You can now add a `WebAuthn Register Passwordless` as the required action for newly registered users.`

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -12,22 +12,17 @@ NOTE: Whether WebAuthn's operations succeed depends on a user's WebAuthn support
...
 - Open the `Authentication -> Required Actions` tab.
 - Click `Register`.
 - Select `Webauthn Register` as `Required Action`.
-- Mark `Enabled` and `Default Action` checkbox.
+- Mark `Enabled` checkbox. Optionally mark `Default Action` checkbox if you want all new created users to be required to register WebAuthn credential.
 
+[[_webauthn-authenticator-setup]]
 ===== Adding WebAuthn Authentication to a Browser Flow
 
 * Select a realm, click on Authentication link, select the "Browser" flow
```
iankko: The Authentication word above should be probably rendered with background too, IOW change it to _click on the `Authentication` link, select the "Browser flow"_

Currently the sentence is rendered like this:
authentication

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -43,6 +38,8 @@ If you want to have WebAuthn required for all users:
 
 image:images/webauthn-browser-flow-required.png[]
 
+* In the `Bindings` menu, change the browser flow to `WebAuthn Browser`
```
iankko, line 41: The "Save" button in the previous sentence (_Select `WebAuthn Authenticator` using the dorp down and click "Save"_) should be probably rendered with background too. Now it looks like this:

webauthn_authenticator_save

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -43,6 +38,8 @@ If you want to have WebAuthn required for all users:
...
+
```
iankko, line 42: The _* Set the alias to "Conditional 2FA"_ sentence below is missing 'Click "Save" button" part' (it's not a problem of this PR though, looks the change was introduced before).

Right now it looks like:
conditional_2fa_no_save

But would be good (IMHO) if it would read as follows:
_Set the alias to "Conditional 2FA" and click on `Save`_

iankko: Yet (from the testing), the _* Select `Condition - User Configured` using the drop down and click on "Save"_ sentence below is missing the sentence, that the requirement of the _Select `Condition - User Configured` execution should be set to `Required` (defaults to `Disabled`). Thus the proposal is to change the sentences as follows (the following image displays how it looks currently):

user_configured

The change would be to add the sentence _Set the Requirement of `Condition - User Configured` execution to Required` so the flow part looks like this:

webauthn_browser_forms_webauthn_only_if_already_configured

because without changing the condition of `Condition - User Configured` from `Disabled` (the default) to `Required` the user is not asked to enter / login via WebAuthn even when they have WebAuthn already configured.

P.S.: Updated the comment & pasted the proper screenshot, since pasted a wrong one in the original comment.

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+need to configure them in the admin console in the tab `WebAuthn Passwordless Policy`. You can configure this policy as you want, however
+typically the requirements for the security key will be stronger than for the two-factor policy. For example the `User Verification Requirement` can
+be set to `Required` when you configure the passwordless policy.
+
+* Finally configure the authentication flow. Let's assume that we will use same flow called `WebAuthn Browser` as described
+<<_webauthn-authenticator-setup, above>>, but
+we will configure it as follows:
+
+** The `WebAuthn Browser Forms` subflow will contain `Username Form` as the first authenticator. This setting means that the user
+will provide just his or her username as the first step.
```
iankko: Unsure if this is clear enough (given the user was previously instructed how to configure WebAuthn). But this also say _Delete the default `Username Password Form` authenticator_ and add the `Username Form` authenticator instead?

The `WebAuthn Browser` flow (copy of Browser one), user was instructed in previous steps to create to looks like this:

username_password_form

but in the final image, it needs to look like:

username_form

From the above description is it clear, the user needs to remove the `Username Password Form` one, and add the `Username Form` one (since those are different), or should we describe this exactly?

WDYT?

server_admin/topics/authentication/webauthn.adoc
```diff
@@ -167,27 +166,61 @@ A new user carries out the following operations :
...
+If `WebAuthn Authenticator` is set up as required as shown in the first example, then when existing users try to log in,
+they are required to register their WebAuthn authenticator automatically :
 
 - Open the login form.
 - Fill in items, click `Save` and  click `Login`.
 - When the users log in, they are required to register their WebAuthn authenticator.
 - After successful registration, the user's browser asks the user to enter the text as their just registered WebAuthn authenticator's label.
 
-===== View Registered WebAuthn Authenticator
+==== Passwordless WebAuthn
```
iankko: Should this say
 _==== Passwordless WebAuthn together with WebAuthn Two-Factor_, rather than just

 _==== Passwordless WebAuthn_?

 (because it's configuring both of them AFAICT)

server_admin/topics/authentication/webauthn.adoc
iankko: And could we have description / image of sole
_==== Passwordless WebAuthn_ configuration? (the section listed above is describing the combination WebAuthn Passwordless together with WebAuthn Two-Factor AFAICT)

Would the proper flow configuration if wanting only `WebAuthn Passwordless` look like this;?

webauth_passwordless

WDYT?

server_admin/topics/authentication/webauthn.adoc
mposolda: Will fix it. Thanks

server_admin/topics/authentication/webauthn.adoc
mposolda: Will fix it. Thanks

server_admin/topics/authentication/webauthn.adoc
mposolda: Will fix it. Thanks. I've used "final" instead "full" .

server_admin/topics/authentication/webauthn.adoc
mposolda: Thanks, it is not specific to registered users. So I will use:
You can now add `WebAuthn Register Passwordless` as the required action to some user, already known to {project_name}, to test this.

server_admin/topics/authentication/webauthn.adoc
mposolda: Will fix it, Thanks

server_admin/topics/authentication/webauthn.adoc
mposolda: Good point. I will fix this and few other occurences of the "Save" and will use `Save` instead, which will use the background style.

server_admin/topics/authentication/webauthn.adoc
mposolda: Thanks. I've used "Passwordless WebAuthn together with Two-Factor" .This is due the main point of this example is to introduce "Passwordless WebAuthn" as the authentication method and the two-factor is here just because it was configured previously. It doesn't matter which two-factor mechanism exactly is used. The fact that WebAuthn is used as two-factor is just because it was easier to extend this section from previous example.

server_admin/topics/authentication/webauthn.adoc
mposolda: Thanks, I will apply both suggestions from your comment above

server_admin/topics/authentication/webauthn.adoc
mposolda: yes, ok. I've added the sentence like this:
Delete the default `Username Password Form` authenticator and add the `Username Form` authenticator instead.

server_admin/topics/authentication/webauthn.adoc
mposolda: Yes, but I think that `WebAuthn Passwordless` will be usually used together with two-factor as an alternative and not sure if there is much point of adding another configuration example. It should be clear enough how it works from the previous example and too much examples makes the docs more complex (and refactoring harder etc) :) So my preference would be to not introduce more example. Are you fine with that?

mposolda:
@iankko Thanks for the feedback and for the suggestions. I've applied most of your suggestions. Added comments to each of the suggestion, so you can take a look for the details. Please take a look at the latest commit and see if it is fine with you. We can then squash it to single commit (I've used separate commit just for the easier review).

server_admin/topics/authentication/webauthn.adoc
iankko: Sure, np with that. Since `WebAuthn Passwordless` is already partially covered in the previous example, should be possible the user to follow the example from the `WebAuthn Passwordless together with Two-Factor` example.

server_admin/topics/authentication/webauthn.adoc
iankko: Maybe `There is also a separate required action...` was meant by @andymunro ?

iankko:
> @iankko Thanks for the feedback and for the suggestions. I've applied most of your suggestions. Added comments to each of the suggestion, so you can take a look for the details. Please take a look at the latest commit and see if it is fine with you. We can then squash it to single commit (I've used separate commit just for the easier review).

@mposolda Thank you for all the updates! The `server_admin` guide compiles correctly & relevant section looks OK AFAICT.

This (the latest) versions LGTM -- but it doesn't include issues raised by @andymunro yet (IOW treat those separately).

server_admin/topics/authentication/webauthn.adoc
mposolda: @iankko ah, I think that makes sense :) I've changed this sentence already to `There is a separate required action...` . Is it ok to keep it that way or should I change it to `There is also a separate required action..` . WDYT @andymunro @iankko ?

iankko: @mposolda `There is a separate required action ....` looks good enough to me.

@andymunro Kind request - could you please review this one with priority & point out the points, that still need to be addressed we to be able to merge this one? People are already asking for this change (see the related code PR for comments).

TYVM! Jan

mposolda:
@iankko Thanks! I've already addressed most of the issues, which @andymunro suggested. I did it even before your review. Now I resolved all the comments bringed by you and @andymunro , but I just left the 3 comments of @andymunro opened. As mentioned in the comment above, I am still not 100% clear what exactly to do with them.

@andymunro Do you please have a chance to look at the remaining 3 unresolved comments and clarify them? Also do you please have a chance to re-review? Thanks a lot!

mposolda:
@andymunro As pointed above, I think your concerns were addressed, so merged this. If there is any additional work/polishing needed, feel free to create additional JIRA for me or send PR. Thanks!

PR #820, opened 2020-01-15
KEYCLOAK-9383 Fixing broken link to EAP Configuration Guide

andymunro:
@stian Can you check this fixed the broken link?

### DIFF
--- a/server_installation/topics/clustering/securing-cluster-comm.adoc
+++ b/server_installation/topics/clustering/securing-cluster-comm.adoc
@@ -3,4 +3,11 @@
 
 When cluster nodes are isolated on a private network it requires access to the private network to be able to join a cluster or to view communication in the cluster. In addition you can also enable authentication and encryption for cluster communication. As long as your private network is secure it is not necessary to enable authentication and encryption. {project_name} does not send very sensitive information on the cluster in either case.
 
-If you want to enable authentication and encryption for clustering communication see link:https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.0/html/configuration_guide/configuring_high_availability#securing_cluster[Securing a Cluster] in the _JBoss EAP Configuration Guide_. 
+ifeval::[{project_product}==true]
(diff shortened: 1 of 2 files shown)


andymunro:
@Stian Can you check this fixed so we can merge this change?

stianst (changes requested):
This will break for community as it's missing appserver_config_guide_link variable in community. It's also hardcoding the name to JBoss EAP guide, which needs to be WildFly in community.

stianst (changes requested):
Should not use if blocks, but rather just templated links and names for the links. See https://raw.githubusercontent.com/keycloak/keycloak-documentation/0c427ffea17bf46e6ecfd3a8722b53ec07573b3b/server_installation/topics/network/bind-address.adoc as an example.

andymunro:
@stianst You are right. I could set up a variable for Securing a cluster with different definitions in the community versus product template files.  However, on further inspection, I don’t see this topic covered in the WildFly documentation.  I looked in the HA guide which has some of the same content as the EAP Config guide, but it's not there. It's not in the Security Guide either. Do you know if it exists somewhere?

andymunro:
@stianst This is where I would expect to find securing a cluster information for WildFly: https://docs.wildfly.org/18/High_Availability_Guide.html#JGroups_Subsystem

I don't see it anywhere.  The reference from the SSO guide is not a requirement.  Do you think we need it for the Keycloak version?

stianst:
@sebastienblanc do you know where community docs for securing JGroups on WildFly can be found?

sebastienblanc:
@stianst TBH I have no idea :) but I can try to ask around

stianst:
@sebastienblanc hehe, wrong Sebi ;)

@slaskawi how about you? Do you know any community docs for enabling security on JGroups?

andymunro:
@stianst, @slaskawi I have not had any feedback.

In the SSO Guide, the reference to the EAP guide is not to a required procedure.  I propose that we include the reference for SSO and exclude it for Keycloak.

slaskawi (approved):

andymunro:
Thanks for approving this @slaskawi.  However, the change I made points to the WildFly High Availability Guide (https://docs.wildfly.org/18/High_Availability_Guide.html#JGroups_Subsystem) for people who want to enable authentication and encryption for clustering communication.  However, I don't see that information there, so I think I should remove that reference.

stianst (approved):

PR #828, opened 2020-01-28
Add missing filters to XML configuration example

alborrajo:
Add filters that are present through the annotations but not in the XML configuration example. These filters are needed to obtain the users' authorization.


abstractj:
@pedroigor could you please review?

PR #844, opened 2020-02-10
KEYCLOAK-9632 Release notes for improvements to locale handling

stianst:


stianst:
Code PR: https://github.com/keycloak/keycloak/pull/6743

release_notes/topics/9_0_0.adoc
```diff
@@ -21,6 +21,28 @@ available authentication mechanisms for two-factor authentication. It should be
 considering the fact that user can have more OTP or WebAuthn credentials. There is also better support for passwordless WebAuthn authentication.
 Finally, we did some work on defects related to the authentication flows.
 
+== Improved handling of user locale
```
mposolda, line 24: The content is good, but it seems that this algorithm for selecting locale is now described just here in the release notes? I wonder that release notes is probably something, which users read just in case that they upgrading to newer version. Would it be slightly better to describe this in some other dedicated section and just reference this section from the release notes instead?

stianst: Moved to server admin guide, and included reference both in release notes and upgrade guide

mposolda (approved):
@stianst Thanks!

PR #847, opened 2020-02-18
KEYCLOAK-12960 Add docs for deprecated methods in JsonWebToken

stianst:


upgrading/topics/keycloak/changes.adoc
```diff
@@ -9,6 +9,14 @@ is updated for a user.
 
 See the link:{adminguide_link}#_user_locale_selection[{adminguide_name}] for more details.
 
+==== Deprecated methods in token representation Java classes
+
+In 2038 an `int` is no longer able to hold the value of seconds since 1970, as such we are working on updating these to
```
mposolda: Small nitpicking: "In the year 2038" might be more explicit as people may not be sure what exactly number "2038" refers to? I am not sure about this one as not native speaker. So rather approving and leave to you if current formula is ok or not :)

mposolda (dismissed):

PR #851, opened 2020-02-24
KEYCLOAK-12958 Preview feature profile for WebAuthn

mabartos:


server_admin/topics/authentication/webauthn.adoc
```diff
@@ -4,7 +4,9 @@
 
 {project_name} provides the limited support for https://www.w3.org/TR/webauthn/[W3C Web Authentication (WebAuthn)]. {project_name} works as a WebAuthn's https://www.w3.org/TR/webauthn/#rp-operations[Relying Party (RP)].
 
-IMPORTANT: Please note that WebAuthn support is still in development and not yet complete, so we recommend that you use this feature experimentally. Also, this support's specification and user interfaces may change.
```
mposolda, line 7: As I mentioned to you in the private chat, we are in the end going to have WebAuthn enabled by default in the project, but marked as preview in the product. So is it possible to change the docs according to it?

We can possibly keep the previous paragraph "IMPORTANT" just for the project and for the product use the block you added with ":tech_feature_name: WebAuthn" etc?

It is possible to ask somehow in the docs whether you are rendering project and product. Hopefully you are able to figure that and let me know just if not :)

server_installation/topics/profiles.adoc
```diff
@@ -60,6 +60,11 @@ endif::[]
 |Upload scripts through the {project_name} REST API
 |No
 |Deprecated
+
```
mposolda, line 63: s it possible to change this part based on whether we use project (enabled by default) or product (preview) ?

upgrading/topics/keycloak/changes.adoc
```diff
@@ -1,5 +1,13 @@
 == Migration Changes
 
+=== Migration to 9.0.1
```
mposolda: Is it possible to remove this? We're still keeping the WebAuthn in the project enabled by default as mentioned above.

mposolda (changes requested):
@mabartos Thanks for the PR. There will be some changes needed due the fact that we're enabling WebAuthn by default in the project, but leaving as "preview" in the product. See inline comments for the details.

andymunro (dismissed):

server_admin/topics/authentication/webauthn.adoc
mabartos: Ok, thanks.

server_installation/topics/profiles.adoc
```diff
@@ -60,6 +60,19 @@ endif::[]
...
+|web_authn
+|W3C Web Authentication (WebAuthn)
+
+ifeval::[{project_community}==true]
+|Yes
+|Experimental
```
mabartos, line 69: @mposolda I don't know, if the _Supported level_ can be `Experimental` for project?!

mposolda: @mabartos I think it is fine and I don't have any better idea. So i would keep it like that :)

mposolda: @mabartos I think it is fine and I don't have any better idea. So i would keep it like that :)

mposolda (approved):
@mabartos Approving, Thanks

PR #858, opened 2020-03-03
Mention SAML protocol mappers as JavaScript provider consumers

thomasdarimont:


thomasdarimont:
Just noticed that this is currently only supported for OIDC Mappers...
https://github.com/keycloak/keycloak/blob/master/wildfly/server-subsystem/src/main/java/org/keycloak/subsystem/server/extension/ScriptProviderDeploymentProcessor.java#L63

PR #859, opened 2020-03-03
Updated section on the `Credentials` tab

seanmmccormick:
Previously, the process for viewing the credentials tab left out an important step. Namely, that the Credentials tab does not show up until you've saved the client with the Access Type set to confidential. 

This patch adds a sentence to help guide readers to get to the Credentials tab.


mposolda (approved):
@seanmmccormick Thanks for this!

PR #863, opened 2020-03-04
KEYCLOAK-9346 Docs for removal of JS promiseType

stianst:


stianst:
Code: https://github.com/keycloak/keycloak/pull/6665

PR #870, opened 2020-03-05
KEYCLOAK-13168 correcting URL for API documentation

andymunro:
@abstractj  Would you be able to check this link now works?

### DIFF
--- a/topics/templates/document-attributes-product.adoc
+++ b/topics/templates/document-attributes-product.adoc
@@ -36,7 +36,7 @@
 :adminguide_clearcache_name: Clearing Server Caches
 :adminguide_clearcache_link: {adminguide_link}#_clear-cache
 :apidocs_name: API Documentation
-:apidocs_link: {project_doc_base_url}/api-documentation/
+:apidocs_link: {project_doc_base_url}/api_documentation/index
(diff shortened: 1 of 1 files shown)


abstractj (commented):
@andymunro it seems to me that the issue still persists. Here's what I did:

```
mvn clean install -Dproduct -DskipTests=true
firefox target/server_development/master.html
```
At `keycloak-documentation/target/server_development/master.html#admin-rest-api` and tried to click in the links there, they point to https://access.redhat.com/documentation/en-us/red_hat_single_sign-on/7.4/html-single/server_administration_guide/, which returns 404.

Please let me know if there is anything I need to do on my end.

andymunro:
Thanks, @abstractj for your quick reply.  To test this you need to replace 7.4 in the URL with 7.3.  We have not built the 7.4 documentation yet to be available on the Red Hat portal.  Does that make sense?

andymunro:
@abstractj Does this work if you use 7.3 in the URL?

stianst (approved):

andymunro:
Thanks, @stianst I will clone this one and the API broken link one to backport it to 7.3.

PR #873, opened 2020-03-06
KEYCLOAK-13248 Revising cross DC graphic to use standard formats.

andymunro:
@tkyjovsk I have included the revised Cross DC graphic that Jess created.  Can you take a look?


stianst (commented):
We already have basis for creating diagrams here https://github.com/keycloak/keycloak-misc. This does not follow this standard. This needs to be discussed with the wider team before basically create yet another "theme" for diagrams.

andymunro:
Thanks, Stian.  this is helpful.  We need to be consistent. 

We have our Doc Sync on Wednesday and our graphic artist, Jess Schaefer, can join the call.  Let's focus on graphics in that call.

stianst:
I think we need to have PNG version in images directory, with a separtae place for the SVG. Browser aren't always great at rendering SVGs, so not sure docs should use SVGs when hosted

stianst (changes requested):
SVG should go somewhere else as now it ends up being copied into distributed documentation without being used. The PNG is the same I assume for Keycloak and RH-SSO so should go into shared images directory instead of duplicated.

andymunro:
The PNGs contain icons labelled either Keycloak or RH SSO, so we need to keep them separate.

We can store the SVG elsewhere.  How about:  https://github.com/keycloak/keycloak-misc/tree/master/diagrams ?

stianst:
If images are different then we also need to store the different SVGs, as we need to keep the exact same SVG as was used to produce the PNG

SVGs should be stored within this repo. Perhaps add image-sources to the root of this repo?

andymunro:
@stianst Okay, that makes sense.  So are you suggesting I create an images-sources folder at the root of the keycloak-documentation repo?

If so, should both SVGs for Keycloak and RH SSO go there, or should the RH SSO go in the keycloak-documentation-prod repo?

stianst:
Actually thinking some more. Perhaps adding it in a separate place would just make it hard to find. For community docs it wouldn't be a problem to just not copy the SVG into the docs dist, so can just go alongside the PNG. Not sure if that would have any implication on Pantheon/customer portal.

andymunro:
It would not be a problem for Pantheon.  So would you like me to just add the SVG for the RH SSO version in the same place?

stianst:
Yes, please. I can update docs build to filter out SVGs afterwards

andymunro:
@stianst I add two versions of the SVG file in the images directory.  Hope that works.

stianst:
Closing this PR as I've opened a new one here:
https://github.com/keycloak/keycloak-documentation/pull/891

I moved the SVGs so they are in the same location as the PNG, with the same names (except extension). This makes it easier to discover the SVGs.

I've also changed the Keycloak docs build to only include PNGs.

PR #875, opened 2020-03-11
KEYCLOAK-13289 Added redirect_url to Node.js logout

bender316:


securing_apps/topics/oidc/nodejs-adapter.adoc
```diff
@@ -257,6 +257,9 @@ By default, the middleware catches calls to `/logout` to send the user through a
 configuration parameter to the `middleware()` call:
 [source,javascript]
     app.use( keycloak.middleware( { logout: '/logoff' } ));
+    
+When the user-triggered logout is invoked a query parameter `redirect_url` can be passed. 
+This parameter is then used as the redirect url of the OIDC logout endpoint.
```
abstractj: @bender316 thanks for submitting this. My suggestion is to not only include this paragraph, but also a snippet with examples. What do you think?

abstractj (changes requested):
@bender316 thanks for submitting this. My suggestion is to not only include this paragraph, but also a snippet with examples. What do you think?

bender316:
@abstractj I updated the section a bit with an example.

abstractj (approved):

abstractj:
@bender316 thank you. I'm just waiting for Travis, but it looks good to me.

PR #876, opened 2020-03-13
KEYCLOAK-13308 adding steps to set the IPA user password

andymunro:
@hmlnarik Can you verify this correction?


mposolda (approved):
@andymunro Looks good, approving. Thanks for the fix!

I see that travis test failures are unrelated to this (it is external links again).

PR #878, opened 2020-03-18
Fixed image paths in token-exchange.adoc

adnanirfan:
Updated image paths


stianst:
Hi, thanks for the PR, but this is not a valid change. If you look at docs https://www.keycloak.org/docs/latest/securing_apps/index.html#_token-exchange the images are working just fine, and it supposed to use the variable as it is currently, not hardcoded links like what you have added.

PR #879, opened 2020-03-19
KEYCLOAK-13240 Reinstating missing section

andymunro:
@stianst I found the missing section and reinstated it.

### DIFF
--- a/server_installation/topics/clustering/sticky-sessions.adoc
+++ b/server_installation/topics/clustering/sticky-sessions.adoc
@@ -67,3 +67,114 @@ into your `RHSSO_HOME/standalone/configuration/standalone-ha.xml` file in the {p
 
 </subsystem>
 ----
+[[_example-setup-with-mod-cluster]]
+==== Example cluster setup with mod_cluster
+
+In the example, we will use link:http://mod-cluster.jboss.org/[Mod Cluster] as load balancer. One of the key features of mod cluster is, that there is not much
+configuration on the load balancer side. Instead it requires support on the backend node side. Backend nodes communicate with the load balancer through the
+dedicated protocol called MCMP and they notify loadbalancer about various events (eg. node joined or left cluster, new application was deployed etc).
+
+Example setup will consist of the one {appserver_name} {appserver_version} load balancer node and two {project_name} nodes.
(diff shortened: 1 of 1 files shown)


stianst:
I'm not really sure if this section should be reintroduced. Don't know why it was removed in the first place, but I would assume it was intentional.

stianst (changes requested):
We need to identify why this section was removed in the first place. There was probably a reason for it.

andymunro:
Is there a way to see if  PR was issued when this was removed?   Here is the source I found in the history: https://github.com/keycloak/keycloak-documentation/blob/081f1c9903173d3287d1a8e7dcac20a9e6cc61f0/server_installation/topics/clustering/sticky-sessions.adoc
Do you think we should just close this request?

stianst:
So here's the commit it was removed in:
https://github.com/keycloak/keycloak-documentation/commit/692ccf82d63d0512be3b0820ee2d2f063f220ecc

Which leads to this JIRA:
https://issues.redhat.com/browse/KEYCLOAK-8300

@mposolda Can you remember why this section was removed?

mposolda:
@stianst @andymunro Yes, this section was removed intentionally. AFAIR the reasons were discussed and they were:
- mod_cluster is not very used loadbalancer and the mod_cluster project is only in the maintenance mode AFAIR
- We had hard times with make mod_cluster loadbalancer working when upgrading to Wildfly 14. In general, the configuration of mod_cluster subsystem changed quite often and was quite fragile during every Wildfly upgrade. Also we're not testing with mod_cluster, so there was an additional cost with having this configuration up-to-date.

So my vote is to close the KEYCLOAK-13240 as rejected.

We already have some example setup in the clustering chapter, which use the other loadbalancer, so there is no need another section specific to mod_cluster IMO. See https://www.keycloak.org/docs/latest/server_installation/index.html#using-the-built-in-load-balancer . However when looking into this, I see that this example is using domain mode and I am sure that this is the setup, which we don't test. So it's possible that this docs can become out-of-date (or maybe already is out-of-date)... :/

stianst:
Makes sense, closing this PR and the JIRA

PR #885, opened 2020-03-24
KEYCLOAK-10825 Ignoring untrusted peers using LDAPS

slaskawi:
https://issues.redhat.com/browse/KEYCLOAK-10825


stianst (changes requested):
I disagree with adding this - as it's an unsafe option, and my guess is that no-one will use this in development, but will end up doing this as a hack/workaround in production (as mentioned in the issue).

slaskawi:
Closing as per @stianst exmplanation

PR #886, opened 2020-03-25
fix(docs): Change "Reset" to "Set"

ejez:
Change "Reset Password" to "Set Password" in "Creating a New User" docs

Screenshot from 2020-03-25 12-57-02


stianst (approved):

stianst:
Thanks

PR #898, opened 2020-03-27
KEYCLOAK-13560 Correct formatting

andymunro:
@mposolda Can you verify the Create Realm is now bold?


mposolda (approved):
@andymunro Yes, it should be fine. Thanks

PR #901, opened 2020-03-30
KEYCLOAK-13553 qualifying OpenShift procedure as for 3.x

andymunro:
@mposolda Can you verify this?

### DIFF
--- a/openshift/topics/get_started.adoc
+++ b/openshift/topics/get_started.adoc
@@ -71,7 +71,7 @@ $ oc policy add-role-to-user view system:serviceaccount:$(oc project -q):default
 
 ==== Deploying the {project_name} Image using the Application Template
 
-===== Deploying the Template via OpenShift CLI
+===== Deploying the Template via OpenShift 3.x CLI
 
 . List the available {project_name} application templates:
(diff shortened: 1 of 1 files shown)


iankko (changes requested):
Don't agree to add this change. All `oc` commands listed in the documentation are universal (not OCP v3.x vs OCP v4.x dependent).

The only needed difference is the expected form oc `oc login` command at the beginning, which is tracked under [KEYCLOAK-13280](https://issues.redhat.com/browse/KEYCLOAK-13280).

This PR should be closed (not needed).

mposolda:
@iankko @lhanusov Should this PR be closed or not? I am not 100% sure as I see that JIRA KEYCLOAK-13280 contains some additional comments even after there was suggestion to close this PR in the comment from MArch 31. If it should be closed, could you please close it?

lhanusov:
@mposolda Yes, you can cancel this PR. @iankko is right that in the Jira ticket it is described more complexly and this PR does not make sense anymore.

mposolda:
@lhanusov Cool, Thanks!

PR #913, opened 2020-04-14
Update https.adoc

lucas-banali:
The text says: "... can run out of the box without SSL so long as you stick to private IP addresses like` localhost`, `127.0.0.1`,` 10.xxx`, `192.168.xx`, and` 172.16.xx` ... "
If it is a private network it would not be better with IP mask, so the 172.16.x.x network would go up to 172.31.x.x, that is, 172.16.x.x / 12.
So the correction would be:
  "... can run out of the box without SSL so long as you stick to private IP addresses like` localhost`, `127.0.0.1`,` 10.xxx / 8`, `192.168.xx / 16`, and` 172.16.xx` ... / 12 "

### DIFF
--- a/server_installation/topics/network/https.adoc
+++ b/server_installation/topics/network/https.adoc
@@ -8,7 +8,7 @@ This default behavior is defined by the SSL/HTTPS mode of each {project_name} re
 link:{adminguide_link}[{adminguide_name}], but let's give some context and a brief overview of these modes.
 
 external requests::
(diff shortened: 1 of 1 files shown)


stianst:
Defining as IP masks or ranges would probably be better than .x. I'm closing this PR though as it only updates on place, and also defines masks wrongly (it's 192.168.0.0/16, not 192.168.x.x/16).

PR #914, opened 2020-04-16
Update documentation including Sync Mode

Martin-Idel-SI:
Connects to https://github.com/keycloak/keycloak/pull/6705

Updated screenshots to also include sync mode and updated description for mappers and IDP.


server_admin/topics/identity-broker/mappers.adoc
```diff
@@ -22,6 +22,11 @@ image:{project_images}/identity-provider-mapper.png[]
 Select a mapper from the `Mapper Type` list.  Hover over the tooltip to see a description of what the mapper does.  The
 tooltips also describe what configuration information you need to enter. Click `Save` and your new mapper will be added.
 
+The mapper will update user information when the user logs in repeatedly according to the `Sync Mode Override`. Choose `inherit` to
```
mposolda: My vote would be to make the documentation a bit clearer for someone, who doesn't know so much context. I would change the last few sentences to something like:

_Choose `legacy` to
keep the mapper behaviour to keep the behaviour in the old Keycloak versions. Choose `import` to only import data when the user is first created in Keycloak during the first login to Keycloak with particular identity provider. Choose `force` to always update user data during every user login._

Maybe the itemized list will be even better :) Also the word "Keycloak" will be good to replace with "{project_name}" (See other point for more details).

server_admin/topics/identity-broker/mappers.adoc
```diff
@@ -22,6 +22,11 @@ image:{project_images}/identity-provider-mapper.png[]
 Select a mapper from the `Mapper Type` list.  Hover over the tooltip to see a description of what the mapper does.  The
 tooltips also describe what configuration information you need to enter. Click `Save` and your new mapper will be added.
 
+The mapper will update user information when the user logs in repeatedly according to the `Sync Mode Override`. Choose `inherit` to 
+use the sync mode configured in the identity provider, all other options will override this sync mode. Choose `legacy` to
+keep the mapper behaviour to keep the old behaviour, `import` to only import data when the user is first created in Keycloak
```
mposolda: The word "Keycloak" in the docs needs to be replaced with some statement like {project_name} . The documentation is used to generate documentation for both "Keycloak" project and "rh-sso" product, so the macro will be replaced automatically during docs compilation. See other places of the documentation for the details.

server_admin/topics/identity-broker/configuration.adoc
```diff
@@ -79,4 +79,8 @@ Regardless of which identity provider you are creating, you'll see the following
 
 |Post Login Flow
 |Authentication flow that is triggered after the user finishes logging in with the external identity provider.
+
+|Sync Mode
+|Strategy of how to update user information from the idp through mappers: When choosing `legacy`, the current behaviour is kept,
```
mposolda: Is it possible to add sentence "See the documentation for Identity Provider Mappers for more details" .

mposolda (changes requested):
@Martin-Idel-SI Thanks for the documentation and for all the screenshots! Nice work! I've added 3 inline comments for adding a bit more details to some sections and replace "Keycloak" with "{project_name}" . But otherwise looks nice!

Martin-Idel-SI:
I updated the documentation - please review the changes again. If necessary, I'll squash and rebase on master.

mposolda (dismissed):
@Martin-Idel-SI Thanks! I am approving this.
I am adding also @andymunro for reviewing this PR. He is our documentation specialist.

stianst (dismissed):

server_admin/topics/identity-broker/configuration.adoc
andymunro: US spelling is "behavior"

Martin-Idel-SI:
@andymunro I updated the spelling to "behavior" and squashed to one commit.

server_admin/topics/identity-broker/mappers.adoc
andymunro: Choose legacy to use the behavior in the previous {project_name} version. Choose import to import only data from when the user was first created in {project_name} during the first login to {project_name} with a particular identity provider. Choose force to update user data at each user login.

andymunro (changes requested):
A few suggestions.

Martin-Idel-SI:
Apparently I was too quick to respond here and the squash destroyed some of the suggestions.
In any case, I updated the PR again with a new commit hopefully catching all suggestions!

Martin-Idel-SI:
@andymunro Hopefully applied all changes. Thanks for reviewing!

stianst (approved):

PR #919, opened 2020-04-20
KEYCLOAK-13858 SAML2 Identity Provider - Support for login_hint 

looorent:
## Motivation

When working with SAML IDPs, it is currently not possible to use provider-specific features such as `login_hint` forwarding. The purpose here is to be able to automatically prefill an IDP login form with a given username.

For instance, this feature is supported by ADFS: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-oapx/a8622e66-2285-43c0-bbb9-abfcecdaed86

Here is the use case scenario we have to solve: https://groups.google.com/forum/#!topic/keycloak-user/hbTf_OyhFCA

More details in JIRA: https://issues.redhat.com/browse/KEYCLOAK-13858

### DIFF
--- a/server_admin/topics/identity-broker/saml.adoc
+++ b/server_admin/topics/identity-broker/saml.adoc
@@ -53,6 +53,9 @@ You must define the SAML configuration options as well.  They basically describe
 |Force Authentication
 |Indicates that the user will be forced to enter in their credentials at the external IDP even if they are already logged in.
 
+|Pass login_hint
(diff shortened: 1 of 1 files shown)


hmlnarik:
Thank you for the well-explained enhancement request. Sadly {{login_hint}} is an OIDC parameter, while SAML has support for this feature by using AuthnRequest's Subject element properly. Therefore this feature request has to be declined.

PR #921, opened 2020-04-22
Update host.adoc

etchepar:
a tiny proposal : the command line reference the old "fixed" provider instead of the default one described above


stianst (approved):
Thanks

PR #931, opened 2020-05-04
Switch ${project.name}, which was not rendering, to {project_name}

razzius:
Currently:

image

This should fix that.


stianst (approved):

stianst:
Thanks

PR #948, opened 2020-05-25
Update resource-overview.adoc

gabrielfalcao:
The usage of word **Scopes** in this section can be confused to with the **Scope** tab at the client configuration level, rather than the **Authorization Scopes** that can be associated with resources.

scopes-scope-authorization-scopes


pedroigor (commented):
Thanks @gabrielfalcao !

PR #955, opened 2020-06-12
index.html change when web-context is changed

shashank-khede:
Whenever the web-context is updated we also should make a change in index.html under the welcome-content folder. If we don't do this change then the server throws 404 Not found when we hit the base URL of the server.

### DIFF
--- a/server_installation/topics/config-subsystem/cli-recipes.adoc
+++ b/server_installation/topics/config-subsystem/cli-recipes.adoc
@@ -18,6 +18,12 @@ For domain mode, this would mean something like:
 ----
 /subsystem=keycloak-server/:write-attribute(name=web-context,value=myContext)
 ----
+_Note: If you change the web-context, change index.html under welcome-content folder as well. For above example it will be_
+[source]
+-----
+window.location.href = "/myContext/"
+<a href='/myContext'>
+-----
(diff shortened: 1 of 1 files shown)


vmuzikar (changes requested):
@shashank-khede Thank you for your contribution! Could you please create a JIRA for this?

I'm not entirely sure this change belongs here. This section of the docs is about CLI and this note describes how to modify a theme. Maybe it's rather a bug and instead of documenting it, we could fix it. :)

@abstractj @douglaspalmer WDYT?

abstractj:
@ssilvert do you think it makes sense?

ssilvert:
> @ssilvert do you think it makes sense?

I don't understand the issue.  I just tried it and the welcome page seems to work fine.

vmuzikar:
@ssilvert I was able to reproduce it before IIRC. I think it's reproducible only when you change the default context, e.g. changing just the host is not enough.

ssilvert:
We need a JIRA with clear steps to reproduce.  If nobody wants to look into it we will just close the PR.

abstractj:
@shashank-khede closing it for now. If you would like to create a Jira with clear steps to reproduce the issue, please let us know.

PR #970, opened 2020-07-02
KEYCLOAK-13837 Admin cli command has alias parameter should be keyAlias

heikkis:
Fixed the documentation as the ticket says

### DIFF
--- a/server_admin/topics/admin-cli.adoc
+++ b/server_admin/topics/admin-cli.adoc
@@ -369,13 +369,13 @@ For example, on:
 +
 [options="nowrap"]
 ----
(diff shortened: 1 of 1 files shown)


ASzc (approved):
Confirmed in [JavaKeystoreKeyProviderFactory.java](https://github.com/keycloak/keycloak/blob/b93a6ed19fa69becca18e07e6e672fadc464497c/services/src/main/java/org/keycloak/keys/JavaKeystoreKeyProviderFactory.java#L46)

@abstractj LGTM

abstractj (approved):

PR #976, opened 2020-07-10
KEYCLOAK-14715 Documentation for the upgrade to Wildfly 20

mposolda:


server_installation/topics/operating-mode/crossdc.adoc
```diff
@@ -323,6 +323,12 @@ Issues related to authorization may exist just for some other versions of {jdgse
 +
 It is currently required to have different configuration files for the JDG servers on both sites as the Infinispan subsystem does not support replacing site names with expressions. See link:https://issues.redhat.com/browse/WFLY-9458[this issue] for more details.
 
+NOTE: The _PUBLIC_IP_ADDRESS_ below refers to the IP address or hostname, which can be used for your server to bind to. Note that
+every {jdgserver_name} server and {project_name} server needs to use different address. During example setup with all hosts on single laptop,
```
iankko: s/During example setup with all hosts on single laptop,.../During example setup with all the servers running on the same host,.../g ?

server_installation/topics/operating-mode/crossdc.adoc
```diff
@@ -323,6 +323,12 @@ Issues related to authorization may exist just for some other versions of {jdgse
 +
 It is currently required to have different configuration files for the JDG servers on both sites as the Infinispan subsystem does not support replacing site names with expressions. See link:https://issues.redhat.com/browse/WFLY-9458[this issue] for more details.
 
+NOTE: The _PUBLIC_IP_ADDRESS_ below refers to the IP address or hostname, which can be used for your server to bind to. Note that
```
iankko, line 326: `Note that every {jdgserver_name} server and {project_name} server needs to use different address.`

Is it really they need to be a different IP address / hostname, or is it different `IP:port` combination enough? If the latter is the case (different IP address / hostname + port combination is enough), I would probably replace `_PUBLIC_IP_ADDRESS_` with `_SOCKET_ADDRESS_`, so it's clear it's just hostname/IP + port combination, which needs to be unique.

More about [socket address term](https://en.wikipedia.org/wiki/Network_socket#Socket_addresses), namely `In this context, sockets are assumed to be associated with a specific socket address, namely the IP address and a port number for the local node, and there is a corresponding socket address at the foreign node (other node), which itself has an associated socket, used by the foreign process. Associating a socket with a socket address is called binding. ` part of it.

server_installation/topics/operating-mode/crossdc.adoc
```diff
@@ -491,6 +511,20 @@ endif::[]
 
 .. The configuration of distributed cache `authenticationSessions` and other caches is left unchanged.
 
+.. It is recommended to add the `remoteStoreSecurityEnabled` property with value `false` (or eventually `true` if you enabled security
```
iankko: (Nitpick) <pre> s/with value `false`.../with the value of `false`.../g </pre>?

upgrading/topics/keycloak/changes.adoc
```diff
@@ -2,6 +2,27 @@
 
 === Migrating to 11.0.0
 
+==== Upgrade to Wildfly 20
+
+The {project_name} server was upgraded to use Wildfly 20 as the underlying container. This does not directly involve any
+specific {project_name} server functionality, but there are few changes related to the migration, which worth mentioning.
```
iankko: (Nitpick) <pre> s/which worth mentioning/which are worth mentioning/g </pre> ?

upgrading/topics/keycloak/changes.adoc
```diff
@@ -2,6 +2,27 @@
...
+Dependency updates::
+  The dependencies were updated to the versions used by Wildfly 20 server. For example, Infinispan is now 10.1.8.Final.
+
+Configuration changes::
+  There are few configuration changes in the `standalone(-ha).xml` and `domain.xml` files. You should follow the <<_install_new_version>>
+  section to handle the migration of configuration files automatically.
+
+Cross-Datacenter Replication changes::
+  * You will need to upgrade {jdgserver_name} server to version {jdgserver_version}. The older version may still work, but it is
+  not guaranteed as we don't test it anymore.
```
iankko: (Nitpick) <pre> s/as we don't test it anymore./as they aren't tested anymore./g </pre> ?

upgrading/topics/keycloak/changes.adoc
```diff
@@ -2,6 +2,27 @@
...
+  * It is recommended to use `protocolVersion` property added to the `remote-store` under configuration of infinispan caches. When
```
iankko: Maybe replace

    It is recommended to use `protocolVersion` property added to the `remote-store` under configuration
    of Infinispan caches.

with

    It is recommended to use the `protocolVersion` attribute of the `remote-store` element when
    configuring Infinispan caches.

?

upgrading/topics/keycloak/changes.adoc
```diff
@@ -2,6 +2,27 @@
...
+  connecting to the {jdgserver_name} server 9.4.18, the recommended version of the hotrod protocol version is 2.9 as the infinispan library
```
iankko: Replace `infinispan` with `Infinispan`

upgrading/topics/keycloak/changes.adoc
```diff
@@ -2,6 +2,27 @@
...
+  version differs among {project_name} server and {jdgserver_name} server. For more details, see the Cross-dc documentation.
```
iankko: Replace `Cross-dc` with `Cross-Datacenter` on both places.

iankko (commented):
Couple of nitpics / comments to clarify yet, but LGTM in overall

iankko:
@andymunro PTAL at the proposed changes too once got a chance

Thanks!, Jan

upgrading/topics/keycloak/changes.adoc
mposolda: Will do. Thanks

upgrading/topics/keycloak/changes.adoc
mposolda: Will do. Thanks

upgrading/topics/keycloak/changes.adoc
mposolda: Thanks. I will use the 
```
It is recommended to use the `protocolVersion` property added tp the `remote-store` element when configuring Infinispan caches.
```
It is technically "property" element, not attribute of the "remote-store" element.

upgrading/topics/keycloak/changes.adoc
mposolda: Thanks! I will use `as they are not tested anymore` . I now recall that Andrew prefers to use something like `are not` instead of `aren't` in the official documentation.

server_installation/topics/operating-mode/crossdc.adoc
mposolda: Will do. Thanks

server_installation/topics/operating-mode/crossdc.adoc
mposolda: For this example, there is _PUBLIC_IP_ADDRESS_ used already and there are no ports involved. Just different virtual hosts like "node1" . I suggest to keep this section as the _PUBLIC_IP_ADDRESS_ is already used in all the example commands below.

server_installation/topics/operating-mode/crossdc.adoc
mposolda: Will update. Thanks

upgrading/topics/keycloak/changes.adoc
mposolda: Will update, Thanks

mposolda:
@iankko Updated per your suggestions. I've added comments to few sections, which I've updated slightly differently than you originally suggested.

server_installation/topics/operating-mode/crossdc.adoc
iankko: Ok, fair enough. Thanks

server_installation/topics/operating-mode/crossdc.adoc
```diff
@@ -491,6 +511,20 @@ endif::[]
 
 .. The configuration of distributed cache `authenticationSessions` and other caches is left unchanged.
 
+.. It is recommended to add the `remoteStoreSecurityEnabled` property with the value `false` (or eventually `true` if you enabled security
```
iankko: I think current

    with the value `false`

should be replaced with

    with the value of `false`.

@andymunro Or is the current form correct, too? WDYT?

mposolda: I've changed to the:

```
with the value of `false`
```
Thanks for the suggestion!

upgrading/topics/keycloak/changes.adoc
```diff
@@ -2,6 +2,28 @@
...
+  The dependencies were updated to the versions used by Wildfly 20 server. For example, Infinispan is now 10.1.8.Final.
+
+Configuration changes::
+  There are few configuration changes in the `standalone(-ha).xml` and `domain.xml` files. You should follow the <<_install_new_version>>
+  section to handle the migration of configuration files automatically.
+
+Cross-Datacenter Replication changes::
+  * You will need to upgrade {jdgserver_name} server to version {jdgserver_version}. The older version may still work, but it is
+  not guaranteed as they are not tested anymore.
+  * It is recommended to use the `protocolVersion` property added tp the `remote-store` element when configuring Infinispan caches.
```
iankko: @mposolda Sorry, but there's a typo (result of a click to right next keyboard key probably) yet. I think it should be `property added to the` instead of `added tp the` current form. WDYT? (sorry for being a way too much pedantic)

upgrading/topics/keycloak/changes.adoc
iankko: This wasn't updated yet (maybe because I marked the comment as hidden?) When updating the point below, could you fix this too?

iankko (changes requested):
Just two very small nitpicks yet, otherwise LGTM in overall

upgrading/topics/keycloak/changes.adoc
mposolda: Updated. Thanks for the catch!

upgrading/topics/keycloak/changes.adoc
mposolda: Updated. Thanks for the catch!

mposolda:
@iankko Updated everything per your last comment. Thanks!

mposolda:
Corresponding codebase PR: https://github.com/keycloak/keycloak/pull/7157

iankko (approved):
@mposolda LGTM, thanks for the updates, Marek!

PR #978, opened 2020-07-16
Fix path-cache lifespan of policy-enforcer

y-tabata:


ASzc (commented):
Doesn't seem to quite align with the code:

- `lifespan` 0 does set the cache `enabled` to `false` in the [constructor](https://github.com/keycloak/keycloak/blob/e9ea1f0e36100304053ee25ca31145a9088d4edc/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/authorization/PathCache.java#L60).

- However, a `lifespan` of `-1` will also set cache `enabled` to `false`, due to the comparison of `this.enabled = maxAge > 0;` When enabled is false, it seems that [no entries can be put into the cache](https://github.com/keycloak/keycloak/blob/e9ea1f0e36100304053ee25ca31145a9088d4edc/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/authorization/PathCache.java#L64), so even though a `-1` lifespan is indeed implemented as infinite, no entries can exist.

I think this is bug, so I've filed [KEYCLOAK-14875](https://issues.redhat.com/browse/KEYCLOAK-14875) This PR may have to wait for that to be merged.

y-tabata:
@ASzc 
Thanks! I've only checked the [`isExpired`](https://github.com/keycloak/keycloak/blob/master/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/authorization/PathCache.java#L174-L176) method.

This should be merged with the code https://github.com/keycloak/keycloak/pull/7288.

ASzc (approved):
The bug PR has been merged, so now this documentation is valid.

@abstractj LGTM

abstractj (approved):

PR #994, opened 2020-07-24
Keycloak 13553

andymunro:
@iankko, @pskopek I updated the OpenShift topic for OCP 4.x, however almost all my changes did not show up.  I emailed you, @pskopek about it.  Will need to follow up on Monday.


andymunro:
Will replace this with a new pull request.

PR #1002, opened 2020-07-29
Update auth-spi.adoc

berendjan:
The secret question required action id is written in lowercase in the example on GitHub, this is confusing for new users as the full code will give a WARN once completed.


ASzc (commented):
I'm not sure what the intended effect is. Your description seems to say the opposite of what your changes are.

Looking at the server code, I see upper case forms being the convention.

PR #959, opened 2020-06-18
KEYCLOAK-14392 Revised Getting Started to fix errors and clarify content

andymunro:
@mposolda, @pdrozd Can you check this PR for 7.4.1?

I personally retested all procedures and made a lot of corrections plus worked with Bin Hu (Terry) a senior TAM who got even further with the final procedure to secure the application.

PR #1007, opened 2020-08-04
[KEYCLOAK-14981] - The repeat steps provided

pedroigor:


pedroigor:
@andymunro It seems steps are repeated after https://github.com/keycloak/keycloak-documentation/pull/959. Could you check I that is the case?

andymunro:
Thanks for approving the other PRs, @mposolda.  Can you check this one also.  It is just removing some duplicate steps.

PR #1009, opened 2020-08-07
KEYCLOAK-15078 Updating version number to 7.4.2

andymunro:
@pskopek @abstractj  I noticed the version number is still 7.4.1

Jira is https://issues.redhat.com/browse/KEYCLOAK-15078  

I already fixed this in downstream.

### DIFF
--- a/topics/templates/document-attributes-product.adoc
+++ b/topics/templates/document-attributes-product.adoc
@@ -1,15 +1,14 @@
 :project_community: false
 :project_product: true
 :project_name: Red Hat Single Sign-On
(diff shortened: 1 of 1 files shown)


sguilhen:
Do we really need to change the prod versions in upstream? Is this template even used for the upstream docs?

andymunro:
@sguilhen I had perhaps wrongly assumed that the upstream needed to corrected for a change Peter asked me to make it downstream.  Either way, this is not needed for 7.4.2.

hmlnarik:
This PR should not be in upstream.

PR #1011, opened 2020-08-10
KEYCLOAK-15106 HTTPS in Docker guide and example

thomasdstewart:
Add some notes about how to get the container to handle https traffic with a real certificate when running with Docker.


thomasdstewart:
https://issues.redhat.com/browse/KEYCLOAK-15106

stianst:
We don't have any documentation for container images in the main docs right now, while we plan to move the docs here as soon as we can, we can't add bits and pieces like this.

PR #1013, opened 2020-08-13
[KEYCLOAK-2940] Documentation for backchannel logout

benjamin37:
This PR provides the documentation related to the oidc backchannel logout. 

It refers to the following PRs in the Keycloak repository:
* https://github.com/keycloak/keycloak/pull/7272
* https://github.com/keycloak/keycloak/pull/7333

Like @pedroigor already suggested we should maybe also extend the documentation https://www.keycloak.org/docs/latest/server_admin/index.html#_oidc and add a logout section there.

### DIFF
--- a/server_admin/topics/clients/client-oidc.adoc
+++ b/server_admin/topics/clients/client-oidc.adoc
@@ -121,6 +121,25 @@ See link:{adapterguide_link}[{adapterguide_name}] for more information.
 To fill in the `Web Origins` data, enter in a base URL and click the + sign to add.  Click the - sign next to URLs you want to remove.
 Remember that you still have to click the `Save` button!
 
+*Backchannel Logout URL*
+
+When {project_name} receives a logout request via `end_session_endpoint` it will post a backchannel logout request at this URL like described in the
+https://openid.net/specs/openid-connect-backchannel-1_0.html[OpenID Connect specification].
+This causes the client to log itself out.
+
+*Backchannel Logout Session Required*
+
+If enabled, the session id is included in the logout token when a backchannel logout request is sent.
+This results in only logging out a specific session. If disabled, all sessions of the corresponding
+user will be logged out.
+
+*Backchannel Logout Revoke Offline Sessions*
+
+If enabled, a `revoke_offline_access` event is added to the logout token. This should result in also logging out offline
+sessions with a backchannel logout request. This event is not specified by the https://openid.net/specs/openid-connect-backchannel-1_0.html[OpenID Connect specification]
+and may have to change in the future. Keycloak itself will respect the `revoke_offline_access` in a logout token when receiving
+a backchannel logout request.
+
 ==== Advanced Settings
 
 [[_mtls-client-certificate-bound-tokens]]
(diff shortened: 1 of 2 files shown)


benjamin37:
Test failures does not seem to be related to my changes.

pedroigor:
@benjamin37 Probably not. As discussed, let's try to write a new logout section with more details on how to use backchannel logout endpoint.

benjamin37:
@pedroigor 

Added a logout section in the oidc part.

The session management part and frontchannel part are not implemented in Keycloak right? 

* Having the scenario that we have a provider (OP) and consumer (RP) and do a login at the consumer via the provider. Triggering a logout at the provider will not result in a logout at the consumer. This should be the case when having the session management part of oidc spec implemented right?
* Having the scenario that we have a provider (OP) and 2 consumers - consumer1 and consumer2 - and do a login at the consumer1 as well as consumer2 via the provider. When triggering a logout at consumer1, only provider and consumer1 are being logged out. The session of consumer2 will still remain.

Should we notice this somehow in the documentation or leave these parts out? Or did I miss something in the configuration of idps/clients so that the logout works via session manamgent or frontchannel logout specification?

pedroigor:
@benjamin37 Documentation looks good to me.

In regards to your bullets above, we should be able to handle both scenarios. It is a proprietary protocol we have for invalidating sessions on clients using backchannel requests depending on how the client is configured.

But yeah, we don't have frontend logout working (as we do for SAML). See https://issues.redhat.com/browse/KEYCLOAK-2939.

benjamin37:
@pedroigor Nice! Should I add a comment in the documentation that the frontchannel logout is currently not implemented in Keycloak or the documentation as is?

Didn't know that there is a configuration for the client to solve the 2 scenarios mentioned in my last comment.
Maybe offtopic for this PR, but: Could you tell me which settings I have to make at the client to solve these scenarios?

pedroigor:
> @pedroigor Nice! Should I add a comment in the documentation that the frontchannel logout is currently not implemented in Keycloak or the documentation as is?
> 
> Didn't know that there is a configuration for the client to solve the 2 scenarios mentioned in my last comment.
> Maybe offtopic for this PR, but: Could you tell me which settings I have to make at the client to solve these scenarios?

Not sure if you need to mention that front-channel is not supported for OIDC. I would say that if not documented, it is not. Up to you though.

Regarding the settings, you should be able to do it by configuring the Management URL to your client and enabling the front-channel logout option.

benjamin37:
> Not sure if you need to mention that front-channel is not supported for OIDC. I would say that if not documented, it is not. Up to you though.

Then I'd leave it like it is.

> Regarding the settings, you should be able to do it by configuring the Management URL to your client and enabling the front-channel logout option.

There is no front-channel logout option in oidc clients, only for saml. Do you mean the "Admin URL" with "Management URL"?

I guess the logic happens in `org.keycloak.services.managers.ResourceAdminManager#logoutClientSessions` where you need to have the Admin URL set and a `client_session_state` note in the client session.

stianst:
JIRA for this: https://issues.redhat.com/browse/KEYCLOAK-18355

pedroigor (approved):

stianst (approved):

stianst:
Fixed merge conflicts here: https://github.com/keycloak/keycloak-documentation/pull/1198

stianst:
Merged the linked PR above which fixes merge conflicts (and squashes to a single commit).

PR #1030, opened 2020-09-11
KEYCLOAK-15158 adjust basic example of Keycloak JS adapter usage

dasniko:
As the JS adapter has evolved, it now tries to add an iframe element to the document body during initialization.
When used like in the basic example as before, there will be an JS error, as the document body is not yet available.
So we've to wait until the document is loaded, thus the initialization is put in a function and the function will be called upon body.onload()

### DIFF
--- a/securing_apps/topics/oidc/javascript-adapter.adoc
+++ b/securing_apps/topics/oidc/javascript-adapter.adoc
@@ -24,17 +24,24 @@ The following example shows how to initialize the JavaScript adapter:
 
 [source,html]
 ----
+<html>
 <head>
     <script src="keycloak.js"></script>
     <script>
-        var keycloak = new Keycloak();
-        keycloak.init().then(function(authenticated) {
-            alert(authenticated ? 'authenticated' : 'not authenticated');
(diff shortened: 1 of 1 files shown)


dasniko:
I don't know if this approach solves all the mentioned errors in KEYCLOAK-15158, but at least one of them is solved this way and it is generally good advice to use the JS adapter in that way.

vmuzikar:
@dasniko Thank you for your contribution. I believe this change is required no more. It was actually a regression (see fix in https://github.com/keycloak/keycloak/pull/7430) and even though it's not best practise to put initialization in `<head>`, it is supported and should be working. WDYT @stianst?

dasniko:
Sad.
The bug was a regression, yes. It might be fixed, ok.
But the example in the docs is not good, as it teaches initialization in `head` tag.
Nevertheless if the JS adapter is working now, the docs should be adjusted to show a better example than the current one. And that's what this PR is for - to give a better example, to make the docs better.
So, IMO this PR is still required.

stianst (approved):

PR #1032, opened 2020-09-13
KEYCLOAK-953: add docs about enabling user to delete account function…

zak905:
…ality


server_admin/topics/users/allow-user-to-delete-account.adoc
```diff
@@ -0,0 +1,46 @@
...
+Allowing the user to delete their account is done through an AIA (Application Initiated Action). You need first of all to enable the action from the admin console. In Authentication menu, go to the Required Actions tab and tick the enabled checkbox for the "Delete Account" action. 
+
+image:{project_images}/enable-delete-account-action.png[]
+
+* Making sure the Account Console client has the `delete-account` role:
+
+The second requirement to enable the delete account functionality is to make sure the "account" client has the `delete-account` role. The `delete-account` role is assigned by default, unless removed by the {project_name} admin. To enable the delete account role, go to Clients menu, and click on the "account" client. 
+Then go to the roles tab, and check if the `delete-account` is among the roles. In case it is not, you can add it by clicking on "Add Role". 
+
+image:{project_images}/delete-account-role.png[]
```
stianst: You can remove this part. delete-account role is an internal role and should not be added/removed by users. If we ever get around to it we will actually make internal clients/roles read-only.

server_admin/topics/users/allow-user-to-delete-account.adoc
```diff
@@ -0,0 +1,46 @@
...
+As stated by the warning message, this action is irreversible, and it implies the deletion of all the user's data in Keycloak. 
+
+Once the user clicks on Delete, he will be prompted to enter his credentials again and redirected to the final confirmation step:
+
+
+image:{project_images}/delete-account-confirm.png[]
+
+
+
+After confirming, the user's account will be deleted.
```
stianst: You're missing the step of adding the delete-account role to users. This should not be added by default, but rather added to those users that should be able to delete their account.

stianst (approved):

MPParsley:
Delete account only seems to be implemented in the Keycloak v2 theme?
If this is so, it should also be documented that one needs to use the Keycloak v2 theme.

zak905:
as far as I know, the v1 theme is being deprecated and there will be only one keycloak theme for the account console. If this is not the case, I can add this info to the docs.

MPParsley:
Thanks @zak905.

The latest Keycloak still offers the base themes for account client. We extend this with our own layout.

Would it be possible to backport the delete account feature for a theme using .ftl templates or to document how one can implement the React "AIACommand" in a regular .ftl theme?

zak905:
I think (but I have not tried) it should be possible to mimic the same functionality in the v1 account console as well by redirecting to /auth/realms/yourrealm/protocol/openid-connect/auth with the required query parameters in addition to `kc_action=delete_action`.  The same requirements also apply (the action should be enabled and the client role should be assigned to the user) 

I have initially implemented this for the old or v1 account console, but after discussions with the maintainers we have decided to remove it and keep it only in the new console. We could have kept it in both for backward compatibility reasons.

MPParsley:
> I have initially implemented this for the old or v1 account console, but after discussions with the maintainers we have decided to remove it and keep it only in the new console. We could have kept it in both for backward compatibility reasons.

Thank you for the info and the tip, is this the code you refer to or is there another commit I can look into? https://github.com/keycloak/keycloak/pull/6265/files#diff-2c92c3123b14440602452ecdab6bf138f249edba5bb3ad5500d53f17557a97eb

zak905:
Right on. Most of the logic is in AccountFormService.java

MPParsley:
> We could have kept it in both for backward compatibility reasons.

Thank you @zak905, upgrading to the v2 client is not an option (would require a lot of theming rework as we extended the base theme and v2 isn't using ftl templates that can be altered).
Could you indicate the steps needed in order to backport this feature to the base theme?

> it should be possible to mimic the same functionality in the v1 account console as well by redirecting to /auth/realms/yourrealm/protocol/openid-connect/auth with the required query parameters in addition to kc_action=delete_action.

I tried adding a static link but it isn't sufficient (luckily or there'd be a csrf issue):
```/auth/realms/{realm}/protocol/openid-connect/auth?kc_action=delete_account```

V2 adds these _query parameters_, which ones are _required_ for the delete to work, and can I get them in the .ftl template without patching Keycloak (AccountFormService.java)?

- client_id=account-console
- redirect_uri
- state
- response_mode=fragment
- response_type=code
- scope=openid
- nonce
- kc_action=delete_account
- code_challenge=
- code_challenge_method=S256

zak905:
Hi @MPParsley, 

No need to override the `AccountFormService.java`  actually. You can just add something similar to this snippet to one of your account pages templates (for example account.ftl):

````
    <form action="/auth/realms/${realm.name}/protocol/openid-connect/auth" class="form-horizontal" method="GET">
        <input type="hidden" id="client-id" name="client_id" value="account">
        <input type="hidden" id="redirect_uri" name="redirect_uri" value="${url.accountUrl}">
         <input type="hidden" id="response-type" name="response_type" value="code">
         <input type="hidden" id="scope" name="scope" value="openid">
        <input type="hidden" id="kc_action" name="kc_action" value="delete_account">
        <button type="submit" class="${properties.kcButtonClass!} btn-danger ${properties.kcButtonLargeClass!}" name="submitAction" value="Delete Account">Delete Account</button>
    </form>

````

I just tried, it seems to work well. Off course, you need to add the necessary roles, and activate the delete account actions as mentioned in the docs. I hope that helps.

MPParsley:
Hi @zak905, this is awesome, thank you very much! Tested it myself and it seems to work just fine.

It does raise two questions:
- Shouldn't there be an email to notify the user that his account will be deleted?
- The account is put in an inactive state, will it automatically be deleted, or is it only deleted once the user logs in again and confirms deletion?

zak905:
> Shouldn't there be an email to notify the user that his account will be deleted?

I don't see any reason, since the user himself is deleting his own account. In case you still want to send it, you can use a an `EventListenerProvider` (here is an [exampe](https://github.com/zak905/keycloak-api-key-demo/blob/master/api-key-module/src/main/java/com/gwidgets/providers/RegisterEventListenerProvider.java#L36)) and capture the `DELETE_ACCOUNT` event. 

> The account is put in an inactive state, will it automatically be deleted, or is it only deleted once the user logs in again and confirms deletion?

I think you are referring to a soft delete. This is a hard delete and the account will be fully deleted when the user confirms. If the user does not re-authenticate and does not confirm, nothing happens.

MPParsley:
Hi @zak905, thank you for your valuable answer.

I'm mainly focused on the UX flow of the delete process:

1. user logs in
2. user navigates to his account page
3. user clicks delete
4. user gets logged out and here the flow is broken: the account is in an inactive state until the user logs in again and (hard) deletes

The user may never log in again to actually (hard) delete his account. We could implement some sort of automatic batch cleanup **or** send an email to resume the delete flow:

5. user opens the confirmation email
6. user clicks on confirm (login) link in the email
7. user clicks on confirm delete account (or the user cancels account deletion and he reactivates account)

zak905:
Welcome. The motives behind adding a re-authentication before confirming the deletion were more of a protection against accidental deletes because as you kown this is a sensitve action, and implies the deletion of the user data and forever. This was something intentional. For example, when you want to access your github settings, you are requested to re-authenticate. The user is actually not logged out, but rather just promted to re-authenticate. So if you hit your browser back button, you can still access the console  again without authenticating. 

From a UX perspective, I believe that informing the user in advance that he will be prompted to re-authenticate and to confirm the deletion would keep things simpler. If you think that the re-authenticate step is not necessary in your case, maybe using a custom action that extends from  `DeleteAction` could be a solution. You can use the same logic, but just set the maxAuthAge to a value higher than 0 here https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/requiredactions/DeleteAccount.java#L185. Actually, this can made configurable in the future, but for now this is the only way.

PR #1038, opened 2020-09-23
KEYCLOAK-15432 extend import/export SAML metadata

cgeorgilakis:


hmlnarik:
Rejecting as the upstream PR was rejected.

PR #1056, opened 2020-11-08
Add jwks_uri to list of endpoints (Spring OAuth2)

dxmann73:
When configuring Spring OAuth2, we need the jwk-set-uri for the provider. Needed some digging to find it, so I thought I'd add this information here.

### DIFF
--- a/server_admin/topics/sso-protocols/oidc.adoc
+++ b/server_admin/topics/sso-protocols/oidc.adoc
@@ -94,7 +94,9 @@ More info together with example is in <<_service_accounts,Service Accounts>> cha
 
 Here's a list of OIDC endpoints that the {project_name} publishes.  These URLs are useful if you are using a non-{project_name} client adapter to
 talk OIDC with the auth server.  These are all relative URLs and the root of the URL being the HTTP(S) protocol, hostname, and usually path prefixed with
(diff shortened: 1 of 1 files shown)


server_admin/topics/sso-protocols/oidc.adoc
```diff
@@ -94,7 +94,9 @@ More info together with example is in <<_service_accounts,Service Accounts>> cha
 
 Here's a list of OIDC endpoints that the {project_name} publishes.  These URLs are useful if you are using a non-{project_name} client adapter to
 talk OIDC with the auth server.  These are all relative URLs and the root of the URL being the HTTP(S) protocol, hostname, and usually path prefixed with
-_/auth_:  i.e. $$https://localhost:8080/auth$$
+_/auth_:  i.e. $$https://localhost:8080/auth$$.
```
andymunro: In general please replace "i.e" with "that is" and replace "e.g." with "for example"

However in this case, please use "for example"

andymunro (changes requested):
One change requested

dxmann73:
Thanks for the suggestion! Changed as requested.

stianst (approved):

PR #1090, opened 2021-01-19
KEYCLOAK-16843 Make Managing User topics easier to find and reuse

andymunro:
@pskopek Can you suggest who could review this topic?

See https://issues.redhat.com/browse/KEYCLOAK-16843


server_admin/topics/users/proc-allow-user-to-delete-account.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+You enable this capability on the *Required Actions* tab.
+
+.Procedure 
+. Click *Authentication* in the menu.
+. Click the *Required Actions* tab.
+. Select *Enabled* on the *Delete Account* row.
++
+.Delete Account on Required Actions tab
+image:images/enable-delete-account-action.png[]
```
abstractj, line 16: @andymunro I just wonder if we should include the metadata for images into the documentation, considering people with visual impairments. For example:

```
image:images/enable-delete-account-action.png[Delete account on required actions tab]
```

@hmlnarik does it make sense?

andymunro: @hmlnarik @abstractj Yes, it makes sense.  I have done that elsewhere, so thanks for catching that omission.

I created a checklist of standards based on the IBM style guide and we do want to add that metadata. I added that checklist to the JIRA, but here it is now: https://docs.google.com/document/d/1Ei2YTokkh7VfO4AwospoSRzHXjFQ_GYCKkfELSpbYYY/edit#heading=h.99n2xdpeusco

Note that some standards are different from current usage. For example, IBM guidelines are to use asterisks, not back ticks for UI elements. We need to gradually change our style

PR #1094, opened 2021-01-27
Prevent leaks with SessionRegistryImpl

frantello:
In the way that the definition of sessionAuthenticationStrategy is documented, SessionRegistryImpl is not reciving the sessionDestroyedEvent and is not deleting SessionInformation related to the closed session. If your aplicatión is runing long time or with hight concurrency, it would produce a fast memory leak.

SessionRegistryImpl implements ApplicationListener<SessionDestroyedEvent>, it's must be  registered with a Spring ApplicationContext as a bean to properly recieve the session destroyed event and remove the SessionInformation. Otherwise the listener will not work, never will be invoked and SessionInformation will stay in memory until the application runs.

### DIFF
--- a/securing_apps/topics/oidc/java/spring-security-adapter.adoc
+++ b/securing_apps/topics/oidc/java/spring-security-adapter.adoc
@@ -55,9 +55,27 @@ While its use is not required, it greatly simplifies your security context confi
 ----
 
 
+@Configuration
+public class SessionConfig
+{
+    /**
+    * Defines the session registry as a bean.
+    */
+    @Bean
+    public SessionRegistry sessionRegistry() {
+        return new SessionRegistryImpl();
+    }
+}
+
 @KeycloakConfiguration
 public class SecurityConfig extends KeycloakWebSecurityConfigurerAdapter
 {
+    /**
+    * Injects the session registry bean.
+    */
+    @Autowired
+    private SessionRegistry sessionRegistry;
+    
     /**
      * Registers the KeycloakAuthenticationProvider with the authentication manager.
      */
@@ -72,7 +90,7 @@ public class SecurityConfig extends KeycloakWebSecurityConfigurerAdapter
     @Bean
     @Override
     protected SessionAuthenticationStrategy sessionAuthenticationStrategy() {
-        return new RegisterSessionAuthenticationStrategy(new SessionRegistryImpl());
+        return new RegisterSessionAuthenticationStrategy(sessionRegistry);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


pedroigor (approved):

frantello:
Boring

mhajas (changes requested):
Thank you for the PR @frantello. Could you add some Jira? Is this the same issue as [this](https://issues.redhat.com/browse/KEYCLOAK-15530)? If yes, I was not able to reproduce the memory leak you described. Could you please let me know if I did something wrong in [this comment](https://issues.redhat.com/browse/KEYCLOAK-15530?focusedCommentId=15494136&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15494136)?

frantello:
Hi @mhajas, thanks for your support.

After read the Issue that you mean, I must say, yes look related, but not exactly the same thing.

This PR don't talk about single sign out, it's not a must in a single sign on context, some apps would require a live session after a eventual global logout.

If we stick in that scenario. If you define the SessionRegistryImpl out of Spring Context, in the way that documentation suggests, It will never receive the logout event, and SessionInformation will remain in that registry until the application restarts, what I or any tool like Eclipse Memory Analyzer identifies as a leak.

The way to go if you want to reproduce, always without single sign out, just add a break point over the event handler and check that it's never called after session expiration or local logout. You can repeat many times, and for every time you do, you will get an extra SessionInformation in the bag with the others still remaining there, in other words, your SessionRegistryImpl will grow more and more based on new and never deleted SessionInformations.

Systems that scale from 0 to infinite may not notice this issue, the leak never kills, but 24x7 systems with high user activity explode really fast.

Annett357:
My PR (https://github.com/keycloak/keycloak-documentation/pull/1282) is basically the same.

Multiple JIRA issues (usually refering to a memory leak) point to the same issue. It can be verified by using a MAT and checking what happens if sessions are closed. The still occupy the heap and will never be garbage collected unless the RegistryImpl is in its own bean.

Please allow for any PR adjusting the documnetation at this place to pass. The issue in the documentation is known since 2015: https://keycloak.discourse.group/t/outofmemory-in-sessionregistryimpl-spring-security-adapter-config/6814

stianst:
@mhajas @pedroigor can we get this issue resolved, and get at least one of the many PRs around this merged?

stianst:
Looks like we should merge https://github.com/keycloak/keycloak-documentation/pull/1282 instead perhaps?

mhajas:
@stianst We can, even though I was not able to reproduce it. I remember it was working with the proposed solution as well. 

I would vote for the https://github.com/keycloak/keycloak-documentation/pull/1282

stianst:
Thanks, I merged https://github.com/keycloak/keycloak-documentation/pull/1282 and closed this one for now

PR #1097, opened 2021-01-29
KEYCLOAK-16945 Adding standard text about metering

andymunro:
dnaro@redhat.com I know this is not what is needed, but it's easier to see where I have made changes.  The main questions is how do include the new content.   This does not work:

include::https://github.com/redhat-documentation/mw-shared-modules/blob/master/modules/ref_runtimes_metering_labels.adoc[]

### DIFF
--- a/release_notes/topics/product/7_4_final.adoc
+++ b/release_notes/topics/product/7_4_final.adoc
@@ -156,6 +156,8 @@ Two features have a change in status:
 
 * Upload of scripts through admin rest endpoints/console is deprecated. It will be removed at a future release.
 
+include::https://github.com/redhat-documentation/mw-shared-modules/blob/master/modules/ref_runtimes_metering_labels.adoc[]
+
 = Fixed Issues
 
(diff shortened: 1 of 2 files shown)


oraNod:
@andymunro Did you rebase your branch? The files are in runtimes-common. 

1. Create a symlink to that directory inside the release notes directory. 
2. Add the attributes file in master.adoc. 
3. Define the two SSO specific attributes. 
4. Include the reference module. 

You also need to set `:context:` somewhere because SSO docs don't use assemblies. If you don't set `:context:` you get this error during build: 

```
xml:id : attribute value runtimes_metering_labels-{context} is not an NCName, line 258, column 53 (line 258)
Unable to parse the AsciiDoc built DocBook XML
```

Changes you need are in this PR: https://gitlab.cee.redhat.com/keycloak/keycloak-documentation-prod/-/merge_requests/6

andymunro:
Superseded by a separate PR.

PR #1108, opened 2021-02-12
fix: typo on kubectl command to get credentials

Ticiano-mw:


vmuzikar (approved):

vmuzikar:
@stianst We're missing JIRA/GH Issue here but not sure it matters in this case.

PR #1109, opened 2021-02-12
KEYCLOAK-17089

Sascha-Block:
Related to KEYCLOAK-17089

### DIFF
--- a/getting_started/topics/sample-app/proc-installing-sample-code.adoc
+++ b/getting_started/topics/sample-app/proc-installing-sample-code.adoc
@@ -37,6 +37,13 @@ $ cd {quickstartRepo_dir}/app-profile-jee-vanilla/config
 ----
 $ mvn clean wildfly:deploy
 ----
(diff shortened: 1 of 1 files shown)


Gamerdachs (approved):

stianst:
Thanks, did you see this issue when using an old version of Java?

We could just set those properties in the parent POM I'd say, rather than documenting a work-around like you've done here.

Sascha-Block:
it should be clarified in the documentations could occur, and yes those properties should be defined in the parent POM

stianst:
Closing this as it should be fixed in quickstarts directly, we shouldn't have a workaround in the docs here.

Sascha-Block:
@stianst Simply closing tickets and issues isn't a solution or even progress for a project.

stianst:
@Sascha-Block this PR is invalid, and the issue it refers to has been closed with a proper fix:

https://github.com/keycloak/keycloak-quickstarts/blob/b1e556e52472e60f527c5610106507e791cc48ac/pom.xml#L73-L74

So, not really sure what the problem is?

PR #1120, opened 2021-02-28
KEYCLOAK-15906

kaja-mohideen:
1/ Updating doc as suggested in the ticket.
2/ Updated an external URL which moved to https to make the tests pass.


mhajas (approved):
Thank you for the PR @kaja-mohideen. The change in the spring xml file look good to me, however, I am not sure about the URL changes. Maybe, we should create a separate PR for that, WDYT @hmlnarik?

hmlnarik (approved):
Approving per mhajas' review

PR #1121, opened 2021-03-01
fix typo (fontawsome, not fortawesome)

fabienhinault:


fabienhinault:
It seams it is really fortawesome. Sorry.

PR #1135, opened 2021-03-12
Update proc-create-realm.adoc

blanchardb:


blanchardb:
There is some incorrect spelling but I was unable to go back and edit my changes.  Please take these changes into consideration.

PR #1136, opened 2021-03-15
KEYCLOAK-16425 Cleanup documentation regarding the logic for selecting locale

mposolda:
@andymunro Do you please have a chance to review this PR?

The description of the logic for selecting locale was present on two places in the documentation. PR removes this duplication to reduce the risk of one place being outdated (which is exactly the reason of KEYCLOAK-16425 as the info in the development guide was outdated). Instead of this, there are links for both documentation sections.


andymunro (approved):
Looks good, Marek

mposolda:
@andymunro Thanks!

PR #1140, opened 2021-03-18
KEYCLOAK-17284 documentation

danielFesenmeyer:
KEYCLOAK-17284: Documentation of the new tabs of the client scopes evaluate screen: "Generated ID Token" and "Generated User Info", updated screenshot


mposolda (dismissed):
@danielFesenmeyer Looks good to me. Thanks for the PR!

@andymunro Do you please have a chance to review this PR?

server_admin/topics/clients/client-scopes.adoc
```diff
@@ -129,11 +129,16 @@ NOTE: If you want to see how you can send a custom value for a `scope` parameter
...
+===== Generating Example Tokens and User Info
 
-To see an example of a real access token, generated for the particular user and issued for the particular client, with the specified
-value of `scope` parameter, select the user from the `Evaluate` screen. This will generate an example token that includes all of the
-claims and role mappings used.
+If you select a user in the `Evaluate` screen, the following example data is generated:
+
+* Generated Access Token: An access token which will be generated and sent to the client when the selected user is authenticated.
+* Generated ID Token: An ID Token which will be generated and sent to the client when the selected user is authenticated.
+* Generated User Info: The information about the user which is returned by the UserInfo Endpoint.
```
andymunro: Use which to indicate extra detail you are giving about an item.  In other words, if you were to leave off that detail, they would still understand what you are talking about.
Use that when you are further defining the item.  In other words, it is not just any ID token, but an ID token that is generated and sent...
So for this section, please use that instead of which.

andymunro (changes requested):
@mposolda Looks good. Just one minor change requested.

server_admin/topics/clients/client-scopes.adoc
danielFesenmeyer: @andymunro Thanks for the review and your detailed explanation. I will adjust this section accordingly.

danielFesenmeyer:
I've updated the PR as requested.

mposolda (approved):

PR #1156, opened 2021-04-20
KEYCLOAK-17824 Fix ReleaseNotes test failure

mhajas:
The failure was caused by the false matching of `_` in adminguide link with `_` around Artifact. I fixed by removing _ around Artifact, if you have any idea how to fix this without removing it please let me know I will update.


hmlnarik (approved):

hmlnarik:
Test failures are unrelated to this PR.

PR #1145, opened 2021-03-29
[KEYCLOAK-16760] [KEYCLOAK-14894] [KEYCLOAK-17433] Keycloak documentation changes related with Wildfly 22 and Wildfly 23 upgrades

iankko:
[KEYCLOAK-14894] Update versions of (various) URN namespaces
    to match their respective Wildfly 22 counterparts
    
    [KEYCLOAK-16760] Update Keycloak documentation for the
    upgrade to Wildfly 22
    
    Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>

<hr/>

    [KEYCLOAK-17433] Keycloak documentation changes related
    with Wildfly 23 upgrade
    
    Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>

<hr/>

_Note:_ Needs to be merged together with:
* [Code change](https://github.com/keycloak/keycloak/pull/7857)
* [Website change](https://github.com/keycloak/keycloak-web/pull/164)

PR #1161, opened 2021-04-29
[KEYCLOAK-17902] Fix failures of 'org.keycloak.documentation.test.ReleaseNotesTest' when building product documentation

iankko:
[KEYCLOAK-17902] Fix failures of 'org.keycloak.documentation.test.ReleaseNotesTest'
    when building product documentation
    
    * Add the two referenced, but (so far) missing files to include,
    * Define the attributes expected by metering labels
    
    Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>

@andymunro @oraNod PTAL once got a chance (pls also see a comment / RFC below)

Thanks,
Jan

### DIFF
--- a/runtimes-common/attributes/runtimes-attributes.adoc
+++ b/runtimes-common/attributes/runtimes-attributes.adoc
@@ -1,0 +1,44 @@
+//
+// This file contains attributes for Red Hat Runtimes product docs.
+//
+
+//
+//Metering labels: product specific
+//
+
+//Each product team must uniquely define two attributes in their projects.
+//This file gets automatically overwritten so treat it as READ only!!
+
+//Define the correct product version.
+//:component-version: x.y.z
+
+//Define the component name.
+//:component-name: "Data_Grid"
+//:component-name: "Vert.X"
+//:component-name: "EAP"
+//:component-name: "JBoss_Web_Server"
+//:component-name: "SSO"
+//:component-name: "AMQ_Broker"
+//:component-name: "Quarkus"
+//:component-name: "Spring_Boot"
+//:component-name: "Thorntail"
+//:component-name: "Node.js"
+
+//Be sure ProductName resolves if you don't already define it.
+//:ProductName: Data Grid
+
+//
+//Metering labels: common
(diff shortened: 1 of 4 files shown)


iankko:
Per [the suggestion](https://github.com/keycloak/keycloak-documentation/pull/1145#pullrequestreview-647338245) from @hmlnarik have split the fix into a separate PR

iankko:
@andymunro @oraNod The commit is related with changes done via KEYCLOAK-16945 and KEYCLOAK-16856 made formerly downstream.

But compared to earlier downstream patches needed to make this change a bit different (conditionally include it only when product documentation is build, and despite the `:component-name:` form in `runtimes-common/attributes/runtimes-attributes.adoc` (suggesting `SSO`) needed to use the full project name form (`Red Hat Single Sign-On`) instead since we abandoned the `SSO` abbreviation some time ago already.

Please review / let me know, if you are OK with the proposed changes (despite the divergence from the suggestion).
Otherwise the product documentation build is now passing.

Thanks, Jan

runtimes-common/attributes/runtimes-attributes.adoc
```diff
@@ -0,0 +1,44 @@
+//
+// This file contains attributes for Red Hat Runtimes product docs.
+//
+
+//
+//Metering labels: product specific
+//
+
+//Each product team must uniquely define two attributes in their projects.
+//This file gets automatically overwritten so treat it as READ only!!
```
iankko, line 10: @oraNod One clarification / question for you yet, if possible. Per the comment above this file is to be treated `read-only`, but there are three attributes required to be defined:

```
:ProductName: {project_name}
:component-name: {project_name}
:component-version: {project_version_base}
```

the documentation to be build / rendered correctly. Can we define those three here (and remove their definition from `maste.adoc` in that case), or do we need to consider the `runtime-attributes.adoc` file as immutable (is there a chance it will get updated on `sync with upstream` on read-only) basis in the future again?

andymunro (approved):

PR #1172, opened 2021-05-10
Update proc-installing-server-community.adoc

learnpy2020:
extracting and installing the tar.gz file using powershell.


stianst:
Closing this as we're no longer using this guide

PR #1174, opened 2021-05-12
KEYCLOAK-17456 adding syntax for Powershell

andymunro:
@sguilhen Can you check this looks good. I simply added a version of a command for Windows powershell.


sguilhen:
@andymunro It looks ok to me, but it looks like we will need another PR to fix the problems with the links (see the two build failures). I know they are not caused by this particular PR but the URLs really seem to have changed.

sguilhen (approved):

andymunro:
Thanks for the quick response, @sguilhen. I see the broken links.  This is something that @pskopek has fixed before.  I am happy to fix it, but I would need help to learn how to do it.

PR #1175, opened 2021-05-14
KEYCLOAK-9725 Fixing cross-references to Server Administration Guide

andymunro:
@sguilhen Can you check this one?

For some reason, the use of #_client_scopes cannot work in both the upstream downstream docs. I had to use conditional clauses for the two occurrences of this cross-reference.

### DIFF
--- a/securing_apps/topics/oidc/java/params_forwarding.adoc
+++ b/securing_apps/topics/oidc/java/params_forwarding.adoc
@@ -17,7 +17,13 @@ and the parameter `scope=offline_access` will be automatically forwarded to the 
 
 The supported parameters are:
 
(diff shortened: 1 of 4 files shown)


stianst (changes requested):
Not sure why, but seems upstream the default anchors include "_" while in downstream it doesn't. Use of conditions like this is not a good approach, and I think the only proper way is to explictily define the anchor.

stianst (changes requested):
Correct fix here is to just add anchor "[[_client_scopes]]" that will add the same anchor both for community and prod

stianst:
Looked into this some more and neither approach are sufficient as there is a large number of broken links due to this problem. Will open a new wider bug report against this. Will send you an email with more details.

PR #1177, opened 2021-05-17
KEYCLOAK-17022 Fix cross-reference in installation guide.

andymunro:
@sguilhen Can you check this one?

### DIFF
--- a/topics/templates/document-attributes-product.adoc
+++ b/topics/templates/document-attributes-product.adoc
@@ -93,7 +93,7 @@
 :installguide_name: Server Installation and Configuration Guide
 :installguide_link: {project_doc_base_url}/server_installation_and_configuration_guide/
 :installguide_clustering_name: Clustering
-:installguide_clustering_link: {installguide_link}#_clustering
+:installguide_clustering_link: {installguide_link}#clustering
 :installguide_database_name: Database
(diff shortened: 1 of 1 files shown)


sguilhen (approved):

stianst:
Link in community is #_clustering, see https://www.keycloak.org/docs/latest/server_installation/#_clustering

andymunro:
@stianst For some reason, the links in the product don't work with the underscore.  It's not just this link.

stianst:
In this PR you are breaking the Keycloak docs to fix the RH-SSO docs though.

andymunro:
@stianst I am leaving the community attributes definition as #_clustering

stianst (changes requested):
I see the link is only changed in prod docs, but it's not very good to have different links in community/prod here as it makes it harder to maintain. Please add [[_clustering]] to the corresponding header to make the link the same in both versions..

stianst:
Here's a correct fix that leaves the link the same for community and product:
https://github.com/keycloak/keycloak-documentation/pull/1202/files

stianst:
This is a wider problem and there are a number of links to other sections that are broken. Closing this issue and created a wider issue here KEYCLOAK-18408.

PR #1179, opened 2021-05-18
KEYCLOAK-16221 Correcting wrong versions of EAP/SSO during upgrade.

andymunro:
@pskopek Can you verify this?


pskopek (approved):
It is correct.

PR #1180, opened 2021-05-20
KEYCLOAK-15932 Adding a sample vault configuration

andymunro:
@hmlnarik Can you check if this works?


hmlnarik:
@sguilhen Could you please check this PR?

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
 The configuration above tells {project_name} to setup the custom Elytron provider and use the key resolver that is created by
 the custom factory.
 endif::[]
+
+=== Sample Configuration
+
+The following is an example of configuring a vault and credential store.  The procedure involves two parts:
+
+* Creating the credential store and a vault, where the credential store and vault passwords are in plain text.
+* Updating the credential store and vault to have the password encrypted by using a mask provided by `elytron-tool.sh`.
```
sguilhen: The password is actually never encrypted, just masked. I would rephrase this to remove any reference to encryption.

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
+
+==== Configuring the credential store and vault without encryption
+
+You create the credential store and a vault where the credential store and vault passwords are in plain text.
+
+.Prerequisites
+
+* A running LDAP instance has `BIND DN credential: secret12`.  
+
+* The alias uses the format <realm-name>_< key-value> when using the default key resolver. In this case, the instance is running in the realm `ldaptest` and `ldaptest_ldap_secret` is the alias that corresponds to the value `ldap_secret` in that realm.
```
sguilhen, line 237: the resolver actually replaces any occurence of underscore character (which is used as the separator between realm and key name) with double underscore characters in both realm and key names. That means that if we have, for example, a key named `ldap_secret`, the final combined key will be `ldaptest_ldap__secret` - notice the final key has replaced the original `_` in the key name with double underscore chars.

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
+
+.Procedure
+
+. Create the credential store.
```
sguilhen: "Create the Elytron credential store"

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
++
+[source,bash,subs=+attributes]
+----
+[standalone@localhost:9990 /] /subsystem=elytron/credential-store=test-store:add(create=true, location=/home/test/test-store.p12, credential-reference={clear-text=testpwd1!},implementation-properties={keyStoreType=PKCS12})
+----
+
+. Add an alias to the credential store.
```
sguilhen, line 250: It might be good to reinforce somehow that the alias must match what the resolver would produce as the final key - i.e. if using the default resolver the final key is realm_key where occurrences of `_` in both realm and key must be replaced by double`_` . Notice how command line bellow is correctly doing that with the 'ldap_secret' part.

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
+----
+
+. Add an alias to the credential store. 
++
+[source,bash,subs=+attributes]
+----
+/subsystem=elytron/credential-store=test-store:add-alias(alias=ldaptest_ldap__secret,secret-value=secret12)
+----
+
+. List the aliases from the credential store.
```
sguilhen: This is optional and serves only as a way to visually inspect the contents of the keystore that is produced by Elytron.

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
+keytool -list -keystore ./test-store.p12 -storetype PKCS12 -storepass testpwd1!
+Keystore type: PKCS12
+Keystore provider: SUN
+
+Your keystore contains 1 entries
+
+ldaptest_ldap__secret/passwordcredential/clear/, Oct 12, 2020, SecretKeyEntry, 
+----
+
+. Configure the vault SPI.
```
sguilhen: "Configure the Vault SPI in Keycloak"

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
+[source,bash,subs=+attributes]
+----
+/subsystem=elytron/credential-store=test-store:add-alias(alias=ldaptest_ldap__secret,secret-value=secret12)
+----
+
+. List the aliases from the credential store.
++
+[source,bash,subs=+attributes]
+----
+keytool -list -keystore ./test-store.p12 -storetype PKCS12 -storepass testpwd1!
```
sguilhen: Minor comment: this assumes the keytool command is being run in the /home/test/ folder as that is the full path of the store that we created above. It might be good to change the -keystore param to the full path to avoid any misunderstandings. So I would change it to -keystore /home/test/test-store.p12 in the above command.

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
+. Configure the vault SPI. 
++
+[source,bash,subs=+attributes]
+----
+/subsystem=keycloak-server/spi=vault:add(default-provider=elytron-cs-keystore)
+
+/subsystem=keycloak-server/spi=vault/provider=elytron-cs-keystore:add(enabled=true, properties={location=>/home/test/test-store.p12, secret=>testpwd1!, keyStoreType=>PKCS12})
+----
++
+At this point, the vault and credentials store passwords are not encrypted.
```
sguilhen: Again, usage of encrypted - I would avoid this sentence. The credential is stored in plaintext form as a SecretKey in the Keystore produced by Elytron, and the password to the store is being used in plaintext form in the Keycloak Vault SPI configuration. What we can do to avoid leaving this password in plaintext is to mask it, but not encrypt it.

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
+
+. In the LDAP provider, replace `binDN credential` with `${vault.ldap_secret}`.
+
+. Test your LDAP connection.
++
+.LDAP Vault
+image:images/ldap-vault.png[LDAP Vault]
+
+
+==== Adding encryption to the credential store and vault
```
sguilhen: What we do is we mask only the password that can open the Elytron credential store (which in this case is a Java KeyStore). No encryption involved.

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
++
+For example:
++
+[source,bash,subs=+attributes]
+----
+elytron-tool.sh mask --salt 12345678 --iteration 123 --secret testpwd1!
+MASK-3BUbFEyWu0lRAu8.fCqyUk;12345678;123
+----
+
+. Store the masked password on the credential store.
```
sguilhen: "Update the Elytron credential store configuration to use the masked password"

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
++
+[source,bash,subs=+attributes]
+----
+/subsystem=elytron/credential-store=cs-store:write-attribute(name=credential-reference.clear-text,value="MASK-3BUbFEyWu0lRAu8.fCqyUk;12345678;123")
+----
+
+. Update the vault password.
```
sguilhen: "Updated the Keycloak vault configuration to use the masked password"

server_admin/topics/vault.adoc
```diff
@@ -216,3 +216,161 @@ To install and use the above custom provider the configuration would look someth
...
++
+[source,bash,subs=+attributes]
+----
+/subsystem=keycloak-server/spi=vault/provider=elytron-cs-keystore:remove()
+/subsystem=keycloak-server/spi=vault/provider=elytron-cs-keystore:add(enabled=true, properties={location=>/home/test/test-store.p12, secret=>”MASK-3BUbFEyWu0lRAu8.fCqyUk;12345678;123”, keyStoreType=>PKCS12})
```
sguilhen, line 346: This is a valid way, but I think it is possible to update the "properties" param without removing the whole provider using 'write-attribute'. Need to check though.

sguilhen (changes requested):

server_admin/topics/vault.adoc
andymunro: Thanks, @sguilhen.  Are you suggesting that I add this as note, change the example to avoid underscores, or something else.

sguilhen: @andymunro I would add a note about it, I think it is good that the example contains underscores.

server_admin/topics/vault.adoc
andymunro: Okay, thanks.  Let me know and I can change it.

andymunro:
@sguilhen Thanks for the quick update. Somehow I don't see your new comments.

server_admin/topics/vault.adoc
andymunro: @sguilhen Do you have more comments on this PR?

sguilhen: @andymunro no, I think it looks good!

sguilhen (approved):

sguilhen:
It looks good to me now, thanks @andymunro !

andymunro:
@hmlnarik Can you approve and merge this?

PR #1183, opened 2021-05-25
KEYCLOAK-18274 document js adapter oidcProvider interface

igord:
When instantiating Keycloak class it is possible to pass `oidcProvider` parameter which allows keycloak JS adapter to work with other IDPs.

### DIFF
--- a/securing_apps/topics/oidc/javascript-adapter.adoc
+++ b/securing_apps/topics/oidc/javascript-adapter.adoc
@@ -62,6 +62,31 @@ var keycloak = new Keycloak({
 });
 ----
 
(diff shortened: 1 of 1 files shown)


stianst:
Thanks for the PR, but I'm going to close it as we aren't quite ready to advertise/advocate the fact that keycloak.js can in theory be used with other IdPs right now. There's quite a lot of things in keycloak.js that are a bit Keycloak specific, we don't have any tests for other providers, and finally we're not really in the business of creating a generic OIDC JavaScript library to be honest.

stianst:
If you want to progress on this we can have a conversation around it through GitHub Discussions.

PR #1189, opened 2021-05-31
fix: add information about npm pkg

wtrocki:
I recieved feedback from customer/users that keycloak client side is widely undocumented, which was surprising to see.
After checking I found out that in fact for users that use npm there is no documentation provided:

https://www.npmjs.com/package/keycloak-js

This is very rare to see in NPM for package that have such large number of downloads. 
Additionally official docs doesn't mention NPMand recommend using js package from server (which is not be the best advice in modern app dev - if you do CDN's etc.).

Additionally, some old bower repo is still 3th result in google and since it has some docs people using it:
https://github.com/keycloak/keycloak-js-bower - worth removing this repository as it is still poping up to users from google and has some docs.

## What is in this PR

Added simple note in the docs for NPM package, so hopefully google will position this documentation

## What we can do in Keycloak 

Create simple README.md file to point out to this documentation so NPM will have proper link for users.

### DIFF
--- a/securing_apps/topics/oidc/javascript-adapter.adoc
+++ b/securing_apps/topics/oidc/javascript-adapter.adoc
@@ -7,6 +7,8 @@ The library can be retrieved directly from the {project_name} server at `/auth/j
 
 A best practice is to load the JavaScript adapter directly from {project_name} Server as it will automatically be updated when you upgrade the server. If you copy the adapter to your web application instead, make sure you upgrade the adapter only after you have upgraded the server.
 
+NOTE: You can also download package from npm: https://www.npmjs.com/package/keycloak-js
+
 One important thing to note about using client-side applications is that the client has to be a public client as there is no secure way to store client
 credentials in a client-side application. This makes it very important to make sure the redirect URIs you have configured for the client are correct and as specific as possible.
 


wtrocki:
CC @mhdawson

slaskawi (approved):

wtrocki:
Attempt to fix problems with missing readme in https://github.com/keycloak/keycloak/pull/8112

wtrocki:
Hi @andymunro 

This is pretty lightweight change that will help others to find package on NPM. 
Also done fix on keycloak but that will require feedback from the team.

andymunro:
@wtrocki Can you give me more detail on what you need?

wtrocki:
> Can you give me more detail on what you need?

https://github.com/keycloak/keycloak-documentation/pull/1189#issue-658143769

andymunro:
Thanks for the quick reply, @wtrocki.   Unfortunately, I have no context for this request.  I am not sure where this ReadMe would exist, what it would say and how it would be delivered. The Keycloak and SSO documentation contains release note, but not a readme.  @slaskawi can you provide any more context?

wtrocki:
@andymunro That is separate keycloak PR. What we really want here is to inform users that package is also on NPM, which is very valuable information. Almost no one uses keycloak.js from keycloak. It needs to be cached somewhere on CDN etc.


https://www.npmjs.com/package/keycloak-js

andymunro:
@wtrocki Thanks again for the quick reply.  I am happy to help, but I need more detail on what you are asking.  For example, when you say " inform users that package is also on NPM," unfortunately I don't know what package you are talking about.

wtrocki:
Screenshot 2021-06-04 at 16 07 51
Screenshot 2021-06-04 at 16 07 34

Text conversations are hard. I just wanted to make sure that we have mutual understanding.

andymunro:
@wtrocki Unfortunately, I still have no context for this request both in terms of creatng and delivering this ReadMe. @slaskawi you suggested that I would be the right person to address this request.  Is there someone else you could recommend?

mhdawson:
@andymunro in the Node.js/Javascript ecosystem, npm )https://www.npmjs.com/) is a common way for distributing/consuming packages.

@wtrocki noticed that that the JavaScript Adapter referenced in the documentation is downloaded 135 thousand times a week from npm, versus loading from the Server as mentioned in the docs.

The change he suggested is to mention npm as as source in addition to getting it from the server.

```
NOTE: You can also download package from npm: https://www.npmjs.com/package/keycloak-js
```

That might not be the context you were looking for, but I thought I'd see if it helps.

andymunro:
Thanks, @mhdawson for the additional detail.  This helps however I still need to know what is precisely the request.  I understand you need a ReadMe.MD file.  Unfortunately, I have limited experience in MarkDown.  

Does the ReadMe need any more information than the following:
NOTE: You can also download package from npm: https://www.npmjs.com/package/keycloak-js

wtrocki:
My suggestion: 

https://github.com/keycloak/keycloak/pull/8112

mhdawson:
Part of the request was just that this PR be landed which adds that one line to the existing key cloak docs.

The other part is that when an npm package is published typically there is a README.md which provides some documentation.  @wtrocki suggestion for the content is in: https://github.com/keycloak/keycloak/pull/8112/files

The latest version of the npm package was published by one of these users with the last 10 days:

keycloak.bot
keycloak-user
stianst

unfortunately none of these give insight on who is doing the publish as some insight into how the publish is done would help to confirm that adding the README.md where @wtrocki suggested would result it being included in the next publish.

andymunro:
@stianst Do you know how to create this ReadMe.MD file?

PR #1202, opened 2021-06-10
KEYCLOAK-17022 Fix link to clustering section

stianst:


stianst:
This is a wider problem and there are a number of links to other sections that are broken. Closing this issue and created a wider issue here KEYCLOAK-18408.

PR #1206, opened 2021-06-16
Specified encoding parameter for module.xml

marwalsch:
The lacking encoding parameter caused the JDBC module not to be found and 1 1/2 work days of headaches.

### DIFF
--- a/server_installation/topics/database/jdbc.adoc
+++ b/server_installation/topics/database/jdbc.adoc
@@ -13,7 +13,7 @@ After you have done this, open up the _module.xml_ file and create the following
 .Module XML
 [source,xml]
 ----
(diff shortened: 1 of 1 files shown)


marwalsch:
```
14:17:27,676 ERROR [org.jboss.as.controller.management-operation] (ServerService Thread Pool -- 50) WFLYCTL0013: Operation ("add") failed - address: ([
    ("subsystem" => "datasources"),
    ("jdbc-driver" => "sqlserver")
]) - failure description: "WFLYJCA0041: Failed to load module for driver [org.microsoft]"
```

These four lines went unseen in a thousands of lines stacktrace including every failed migration subsequent to the connection which could not be established.

PR #1220, opened 2021-06-30
KEYCLOAK-18466 Configure HTTP client timeouts for adapters

mabartos:
JIRA: [KEYCLOAK-18466](https://issues.redhat.com/browse/KEYCLOAK-18466)

### DIFF
--- a/securing_apps/topics/oidc/java/java-adapter-config.adoc
+++ b/securing_apps/topics/oidc/java/java-adapter-config.adoc
@@ -27,6 +27,9 @@ This is what one might look like:
    },
 
    "connection-pool-size" : 20,
+   "socket-timeout-millis": 5000,
+   "connection-timeout-millis": 6000,
+   "connection-ttl-millis": 500,
    "disable-trust-manager": false,
    "allow-any-hostname" : false,
    "truststore" : "path/to/truststore.jks",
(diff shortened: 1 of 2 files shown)


andymunro (dismissed):

securing_apps/topics/oidc/java/java-adapter-config.adoc
```diff
@@ -150,6 +153,19 @@ connection-pool-size::
   This is _OPTIONAL_.
   The default value is `20`.
 
+socket-timeout-millis::
+  Timeout for socket waiting for data after establishing the connection in milliseconds.
+  Maximum time of inactivity between two data packets.
```
mhajas, line 158:
```suggestion
  Maximum time of inactivity between two data packets. A timeout value of zero is interpreted as an infinite timeout.
```

securing_apps/topics/oidc/java/java-adapter-config.adoc
```diff
@@ -150,6 +153,19 @@ connection-pool-size::
...
+  This is _OPTIONAL_.
+
+connection-timeout-millis::
+  Timeout for establishing the connection with the remote host in milliseconds.
```
mhajas, line 165:
```suggestion
  Timeout for establishing the connection with the remote host in milliseconds. A value of zero means the timeout is not used. The default value is zero.
```

mhajas (commented):
Thank you for the PR @mabartos. I added two suggestions with the addition of more details to properties. Also, this will probably require other changes based on the last comment in the `keycloak/keycloak` PR

securing_apps/topics/oidc/java/java-adapter-config.adoc
mabartos: Done

securing_apps/topics/oidc/java/java-adapter-config.adoc
mabartos: It should be the same like `socket-timeout-millis`... There's used `org.apache.http.params.CoreConnectionPNames`, where the parameter has same behavior like the mentioned `socket-timeout-millis`. It should be: _A timeout value of zero is interpreted as an infinite timeout._

hmlnarik: @mabartos Makes sense to keep this the same as in httpclient docs. 
The other point is valid, the default values should be documented similarly to the other options.

hmlnarik (approved):

PR #1187, opened 2021-05-27
KEYCLOAK-15580 Revise Server Administration Guide so topics are findable and reusable

andymunro:
@pskopek, @pdrozd  As discussed at the program call today, here is the revised Server Administration guide to be reviewed after Keycloak 14 is released.

Because of the number of commits, it will be easier to review the built HTML version of the revised version compared to the existing version.

PR #1228, opened 2021-07-13
Corrected an issue with header injection in HeaderMojo.java

adahms:
This PR corrects an issue with header injection in the HeaderMojo.java file.

Today, the processTopics function attempts to inject a custom header into files by identifying the title of the topic and injecting the header after that on several new lines.

Because the test to identify the header uses .contains instead of .startsWith, the function incorrectly identifies the section to inject the header in files where there is a section ID (for example, 'ID="XYZ') or another line that contains the equals sign. As a result, the header is injected between the section ID in the HTML and the actual title, causing the sidebar to appear in the incorrect location.

This PR changes the function to use 'startsWith' instead, correctly identifying the line on which the file's section title appears, and resolving this issue.

Tested using mvn compile.

### DIFF
--- a/header-maven-plugin/src/main/java/sample/plugin/HeaderMojo.java
+++ b/header-maven-plugin/src/main/java/sample/plugin/HeaderMojo.java
@@ -114,7 +114,7 @@ public class HeaderMojo extends AbstractMojo {
             try(PrintStream ps = new PrintStream(new FileOutputStream(out));BufferedReader br = new BufferedReader(new FileReader(f));){
                 for (String l = br.readLine(); l != null; l = br.readLine()) {
                     ps.println(l);
-                    if (l.contains("=")) {
+                    if (l.startsWith("=")) {
                         break;
                     }
                 }


andymunro (approved):

andymunro:
I tested this on two guides and it definitely works.

andymunro:
@abstractj Can you approve this fix.  The Edit/Report bar in the community documentation is sometimes blocking the text.   Here is an example:  https://www.keycloak.org/docs/latest/getting_started/#installing-a-sample-instance-of-keycloak

andymunro:
@stianst I guess this requires your approval now.

andymunro (approved):
@adahms Thanks for creating this PR.  Actually, I needed to include it in a separate PR to make sure that the Keycloak documentation works correctly.  So this PR can bel closed.

stianst:
@andymunro Not sure what you mean you need to include this in a separate PR? Seems like this fixes the issue on its own and should be merged as is.

andymunro:
@stianst In submitting this Server Administration Guide PR (https://github.com/keycloak/keycloak-documentation/pull/1187/files), I needed to include this change for this to work in community.  So this change is already in master.

stianst:
Oki, closing this one then. @adahms thanks

PR #1254, opened 2021-09-29
KEYCLOAK-18934 Note about losing user sessions in a minor upgrade.

andymunro:
@hmlnarik Would you be able to check this small update?  This is the change:

Warning

In a minor upgrade of Red Hat Single Sign-On, all user sessions are lost. All users will have to log in again.


andymunro:
@mposolda Would you be able to review this very small change for me?

mposolda:
@andymunro Thanks for this note

PR #1264, opened 2021-10-04
CIAM-959 Updating release notes to match downstream changes for SSO 7.5

andymunro:
@pskopek  Can you verify this?

### DIFF
--- a/release_notes/topics/product/7_5.adoc
+++ b/release_notes/topics/product/7_5.adoc
@@ -25,16 +25,16 @@ The {project_name} login theme components have been upgraded to PatternFly 4. Pa
 
 Also, the login theme provides a better user experience and you can define an icon for your custom Identity providers. For details, see the link:{developerguide_link}#custom-identity-providers-icons[Server Developer Guide].
 
-== Users can delete their own accounts 
-You can allow users in a given realm to delete their own account through the Account Console. This capability is enabled by the *Delete Account* action in the Admin Console.  
+== Users can delete their own accounts
+You can allow users in a given realm to delete their own account through the Account Console. This capability is enabled by the *Delete Account* action in the Admin Console.
 
-== Identity brokering sync-mode 
+== Identity brokering sync-mode
 With Identity Brokering Sync Mode, you can now control if user profiles are updated on the first login or on every login from an external Identity Provider. You can also override this behavior on individual mappers.
 
 == Client Session Timeout for OpenID Connect / OAuth 2.0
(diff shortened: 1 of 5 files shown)


stianst:
@andymunro LGTM, but needs a rebase

tests/src/test/resources/ignored-links
```diff
@@ -1,3 +1,8 @@
+https://issues.redhat.com/browse/KEYCLOAK-18994
+https://developers.facebook.com/docs/development/
+https://developers.facebook.com/ 
+https://developers.facebook.com/docs/graph-api 
+https://developers.facebook.com/docs/facebook-login/access-tokens/
```
stianst: Why are you adding these links to ignore?

release_notes/topics/product/7_5.adoc
```diff
@@ -0,0 +1,134 @@
+= Overview
```
stianst, line 1: I'm assuming this is a direct copy of the release notes for 7.5, and there are no changes here?

stianst:
@andymunro FIY when you do a push to your branch with updates, without adding a comment to the PR there's no notification sent by GitHub, so it's important to write a comment on what you changed to let folks know it's been updated.

tests/src/test/resources/ignored-links
stianst: I would suggest removing these changes in this PR, and rather open a separate PR to resolve any broken links. In the long run we should rather have a nightly run on GitHub Actions that check all links, and ideally only check new links added in a PR, as its rather problematic that new PRs can't be merged due to already existing links suddenly breaking.

release_notes/topics/product/7_5.adoc
andymunro: Yes, that is correct.

tests/src/test/resources/ignored-links
andymunro: Okay, @stianst.  To be precise, the links are not broken.  I tested them in a built version of the document and they all work, but they are reported as invalid, so that is why I added them to ignored-links.

So do you want me to remove the addition to ignored-links and request approval of the PR with a note they are not broken, then issue a separate PR adding these URLs to ignored-links?

andymunro:
Thanks, @stianst.  I will add more detail to the push in future.

tests/src/test/resources/ignored-links
stianst: Before merging we review the results anyways and make sure broken links are not caused by the PR

Ignoring the links is not the right fix, we need to figure out why valid links are reported as broken instead.

stianst (approved):

PR #1273, opened 2021-10-19
KEYCLOAK-19144 remove "Installing from the command line" chapter

jonathanvila:
https://issues.redhat.com/browse/KEYCLOAK-19144

Remove the chapter suggesting to install from the command line.


vmuzikar (changes requested):
@jonathanvila I'm really sorry, I miscommunicated this. The chapter of course needs to stay, but it needs to be present only in the Keycloak docs and be removed from the RH-SSO docs. Could you please just wrap it in some `if` instead of removing it (`ifeval::[{project_community}==true]`)? Thank you!

vmuzikar (changes requested):
I can still see the chapter present in the RH-SSO docs when I build it locally as:
```
mvn clean install -Pproduct -DskipTests
```

andymunro (approved):

README.md
```diff
@@ -7,15 +7,15 @@ For more information about Keycloak visit the [Keycloak homepage](http://keycloa
 
 
 Contributing to Keycloak Documentation
---------------------------------------
+----------------------------------------
```
vmuzikar, line 10: Just a nitpick (not a blocker). Why to make this change? Seems unrelated (and even in an unrelated file).

README.md
```diff
@@ -7,15 +7,15 @@ For more information about Keycloak visit the [Keycloak homepage](http://keycloa
...
 
 See our [Contributor's Guide](internal_resources/contributing.adoc). The directory also includes a set of templates and other resources to help you get started.
 
 If you want to file a bug report or tell us about any other issue with this documentation, you are invited to please use our [issue tracker](https://issues.redhat.com/projects/KEYCLOAK/).
 
 
 Building Keycloak Documentation
--------------------------------
+---------------------------------
```
vmuzikar, line 18: Just a nitpick (not a blocker). Why to make this change? Seems unrelated (and even in an unrelated file).

jonathanvila:
It was for forcing the build on github, we can remove those commits, but it
will take more time removing them than leaving as this has no impact on the
result

vmuzikar (approved):

PR #1253, opened 2021-09-29
KEYCLOAK-18911 Removing backup feature from Operator documentation

andymunro:
@hmlnarik Would you be able to check that this unsupported feature is removed?

PR #1277, opened 2021-10-20
KEYCLOAK-19549 specify backup is deprecated and external database is default

jonathanvila:
https://issues.redhat.com/browse/KEYCLOAK-19549


server_installation/topics/operator/keycloak-backup-cr.adoc
```diff
@@ -1,5 +1,7 @@
 
 [[_backup-cr]]
+Backup CR is *deprecated* and could be removed in future releases.
```
vmuzikar: This note needs to be probably elsewhere as this file will be removed in #1253. But I'm not entirely sure where.

server_installation/topics/operator/keycloak-cr.adoc
```diff
@@ -29,7 +29,7 @@ ifeval::[{project_community}==true]
  or an Ingress for Kubernetes
 endif::[]
  for the {project_name} cluster.
-* `externalDatabase` - applies only if you want to connect an externally hosted database. That topic is covered in the xref:_external_database[external database] section of this guide.
+* `externalDatabase` - in order to connect to an externally hosted database. That topic is covered in the xref:_external_database[external database] section of this guide. This is enabled by default, only for testing purposes you can disable it and this will install an embedded PostgreSQL database.
```
vmuzikar: Please see https://github.com/keycloak/keycloak-operator/pull/386#discussion_r733772947

vmuzikar: I'd also add somewhere around here a big warning box that using `externalDatabase: false` is not supported in prod envs.

vmuzikar (changes requested):
@jonathanvila Thanks! Added a few comments.

Additionaly, there are a few mentions of "If you have an external database, you can modify the Keycloak custom resource to support it." I think we need to rephrase a bit. It's not an option anymore but a requirement.

server_installation/topics/operator/keycloak-cr.adoc
jonathanvila: I already answered there..... I think we are not breaking anything because the field is mandatory BUT it has a default value.

server_installation/topics/operator/keycloak-backup-cr.adoc
jonathanvila: well, if the file is removed, then , I think no need to include anywhere that is deprecated

jonathanvila: I've added a comment in #1253 . Are we sure we want to remove completly the mention of the backup feature, when the CRD will still exist ?

vmuzikar: It was requested by support that it needs to be removed as it's confusing to customers, which IMHO makes sense.

Or do you think it'd be better to just document it's deprecated and unsupported and leave the backups docs there? I'm open to both options.

jonathanvila: I think that if we have a feature, we need to have documentation about it, if not it looks like an "easter egg".
I would prefer to add "deprecation" and "unsupported" information everywhere but without removing its documentation.

server_installation/topics/operator/keycloak-cr.adoc
jonathanvila: :+1:

jonathanvila: After revisiting the feature, there's no easy way of defaulting to true without breaking changes.
I will fix the comment.

vmuzikar:
Before proceeding with this PR, we need to double check if https://github.com/keycloak/keycloak-documentation/pull/1253 can be closed as it conflicts with it.

I'm also wondering, whether we should add a short separate chapter (maybe something like "Using the operator in production environment") that would highlight the (un)supported state of all affected components at one place. I think we need to be as explicit as possible. Namely:
* Embedded DB is not supported in prod.
* Backup CRD is deprecated and not supported in prod.
* We fully support using the rest of the CRDs in prod, despite the `v1alpha1` version. We do not plan to make any breaking changes in this CRDs version.

@jonathanvila WDYT? Does it make sense?

server_installation/topics/operator/external-database.adoc
```diff
@@ -2,7 +2,7 @@
 [[_external_database]]
 === Connecting to an external database
 
-You can use the Operator to connect to an external PostgreSQL database by modifying the Keycloak custom resource and creating a `keycloak-db-secret` YAML file. Note that values are Base64 encoded.
+You can use the Operator to connect to an external PostgreSQL database by creating a `keycloak-db-secret` YAML file and setting Keycloak CR externalDatabase property enabled (this is the default value). Note that values are Base64 encoded.
```
vmuzikar: Sorry, missed this before.
```suggestion
You can use the Operator to connect to an external PostgreSQL database by creating a `keycloak-db-secret` YAML file and setting Keycloak CR externalDatabase property enabled. Note that values are Base64 encoded.
```

server_installation/topics/operator/keycloak-backup-cr.adoc
```diff
@@ -1,5 +1,7 @@
...
+
 === Scheduling database backups
```
vmuzikar: @jonathanvila Sorry, one more suggestion. After building the docs I realized this note is very subtle and almost looks like it belongs to the previous chapter:
Screenshot 2021-11-01 at 14 15 46


Maybe we could do something like:
```suggestion
[[_backup-cr]]
=== Scheduling database backups

[WARNING]
====
Backup CR is *deprecated* and could be removed in future releases.
====
```

server_installation/topics/operator/external-database.adoc
vmuzikar: I believe this is still valid.

vmuzikar (changes requested):

server_installation/topics/operator/keycloak-backup-cr.adoc
jonathanvila: OK

server_installation/topics/operator/installation.adoc
```diff
@@ -153,3 +153,10 @@ endif::[]
 
 * For more information on OpenShift Operators, see the link:https://docs.openshift.com/container-platform/4.4/operators/olm-what-operators-are.html[OpenShift Operators guide].
 
+[[_operator_prod_usage]]
+=== Using the {project_operator} in production environment
+
+* The usage of embedded DB is not supported in production environment.
```
andymunro: add "a" before "production"
change "prod" to "production"

server_installation/topics/operator/installation.adoc
```diff
@@ -153,3 +153,10 @@ endif::[]
...
+* Backup CRD is deprecated and not supported in production environment.
```
andymunro: add "a" before "production"

server_installation/topics/operator/installation.adoc
```diff
@@ -153,3 +153,10 @@ endif::[]
...
+* We fully support using the rest of the CRDs in prod, despite the `v1alpha1` version. We do not plan to make any breaking changes in this CRDs version.
```
andymunro: change "prod" to "production"

server_installation/topics/operator/keycloak-cr.adoc
```diff
@@ -29,7 +29,7 @@ ifeval::[{project_community}==true]
  or an Ingress for Kubernetes
 endif::[]
  for the {project_name} cluster.
-* `externalDatabase` - applies only if you want to connect an externally hosted database. That topic is covered in the xref:_external_database[external database] section of this guide.
+* `externalDatabase` - in order to connect to an externally hosted database. That topic is covered in the xref:_external_database[external database] section of this guide. Setting it to false should be used only for testing purposes and will install an embedded PostgreSQL database. Be aware that externalDatabase:false is *NOT* supported in prod environments.
```
andymunro: change "prod" to "production"
Not clear what action is what "will install an embedded PostgreSQL database"  - setting it to False?

server_installation/topics/operator/keycloak-cr.adoc
```diff
@@ -243,4 +243,4 @@ Events:
 
 * Once the installation of {project_name} completes, you are ready to xref:_realm-cr[create a realm custom resource].
 
-* If you have an external database, you can modify the Keycloak custom resource to support it. See xref:_external_database[Connecting to an external database].
+* An external database is the supported option and needs to be enabled in the Keycloak custom resource. Only for testing purposes you can disable it. See xref:_external_database[Connecting to an external database].
```
andymunro: Change "Only for testing purposes you can disable it."
TO
"Use this option only for testing and disable it when you switch to a production environment."

andymunro (changes requested):
A few changes, Jonathan.

server_installation/topics/operator/installation.adoc
jonathanvila: you mean "prod" in the chapter key : "_operator_prod_usage" ?

server_installation/topics/operator/installation.adoc
jonathanvila: :+1:

server_installation/topics/operator/installation.adoc
jonathanvila: :+1:

server_installation/topics/operator/installation.adoc
jonathanvila: :+1:

server_installation/topics/operator/keycloak-cr.adoc
jonathanvila: if you set it to false then it won't use an externalDatabase , and will install a PostgreSQL database pod to be used.

server_installation/topics/operator/keycloak-cr.adoc
jonathanvila: :+1:

vmuzikar: I believe this changed the meaning of the sentence. It now implied the external database os only for testing and should be disabled when you switch to a production environment. That is not true. We need the opposite:
`An external database is the supported option and needs to be enabled in the Keycloak custom resource. You can disable this option only for testing and enable it when you switch to a production environment.`

@andymunro Can you please confirm this change is correct language wise?

andymunro: Looks good, @jonathanvila

andymunro (dismissed):

vmuzikar (dismissed):

server_installation/topics/operator/external-database.adoc
```diff
@@ -2,7 +2,7 @@
 [[_external_database]]
 === Connecting to an external database
 
-You can use the Operator to connect to an external PostgreSQL database by modifying the Keycloak custom resource and creating a `keycloak-db-secret` YAML file. Note that values are Base64 encoded.
+You can use the Operator to connect to an external PostgreSQL database by creating a `keycloak-db-secret` YAML file and setting Keycloak CR externalDatabase property enabled. Note that values are Base64 encoded.
```
andymunro: setting Keycloak CR externalDatabase property to enabled

andymunro (changes requested):
Thanks, a few changes needed.

server_installation/topics/operator/external-database.adoc
jonathanvila: ok

vmuzikar (approved):

stianst (approved):

PR #1279, opened 2021-10-21
KEYCLOAK-19595 ORCID Social Identity Provider

cgeorgilakis:


stianst:
See https://github.com/keycloak/keycloak/pull/8618#issuecomment-965012876

PR #1280, opened 2021-10-21
KEYCLOAK-19433 Release Note about deprecating Spring Boot adapter.

andymunro:
@hmlnarik Can you approve this change requested by John Doyle?


stianst (changes requested):
This is being added to the release notes of the previous release, and not the most release. I'm also not sure if the release notes are the best place for this notice as I would imagine people read those when they upgrade and do not return to these.

andymunro:
@stianst I added a note to the securing apps guide also.  This upstream PR is for 7.4.10, but I will clone it for 7.5.1.

stianst:
I would just change this PR to only cover 7.5. I don't think we should be adding deprecating notes to a version that is pretty much out of support anyways. However, if you think we should add it to both, then a single PR updating both 7.4 and 7.5 release notes would be better here.

andymunro:
Thanks, @stianst I can use one PR for upstream but I will first need these clones of downstream 7.5 PRs forward ported to upstream:

- https://github.com/keycloak/keycloak-documentation/pull/1264 (Release Notes)
- https://github.com/keycloak/keycloak-documentation/pull/1260 (Securing Apps)
This is for 7.5.1 which has a code freeze on Friday 11/5.

stianst (approved):

PR #1282, opened 2021-10-28
Update spring-security-adapter.adoc

Annett357:
We could confirm a memory leak due to the SessionRegistryImpl not created as a separate bean and thus not processing SessionDestroyedEvents. As described in the linked topic (https://keycloak.discourse.group/t/outofmemory-in-sessionregistryimpl-spring-security-adapter-config/6814) this resulted in sessions never being destroyed until the applications went unresponsive because the heap was full.


mhajas (approved):
Thanks for the PR @Annett357 LGTM

PR #1289, opened 2021-11-08
KEYCLOAK-19800 Fixing attribute used in API documentation

andymunro:
@hmlnarik Can you approve and merge this minor correction?  The correct link is rest-api, not restapi.  I did a workaround at 7.5, but for 7.5.1, we need to use the actual directory name: rest-api.


hmlnarik (approved):
Approving at @andymunro's request

PR #1290, opened 2021-11-09
declarative-user-profile for read only attributes

bastianccm:
Rename legacy-user-profile to declarative-user-profile in read only attribute documentation.

### DIFF
--- a/server_admin/topics/threat/read-only-attributes.adoc
+++ b/server_admin/topics/threat/read-only-attributes.adoc
@@ -44,7 +44,7 @@ You can add this configuration to your `standalone(-*).xml` files to the configu
 [options="nowrap"]
 ----
 <spi name="userProfile">
-    <provider name="legacy-user-profile" enabled="true">
+    <provider name="declarative-user-profile" enabled="true">
         <properties>
             <property name="read-only-attributes" value="[&quot;foo&quot;,&quot;bar*&quot;]"/>
             <property name="admin-read-only-attributes" value="[&quot;foo&quot;]"/>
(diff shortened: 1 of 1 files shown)


andymunro (dismissed):

bastianccm:
Hey @stianst @andymunro do I have to do anything to get this merged?

andymunro:
@bastianccm The checks are not completing
Have you tried this command: git pull upstream main

bastianccm:
Hi @andymunro the checks are pending, I don't know how I can trigger them? I rebased anyway, hope this helps.
@keycloak-docs-bot reports them to be successful.

andymunro:
Okay, I think then it's up to @stianst to merge it.

pedroigor:
@bastianccm The documentation you are changing is indeed for legacy. Although you can use that property with declarative too.

For the declarative provider, we want people using the read-only capabilities from the user profile configuration and not using the "old" way of setting read-only attributes.

Does it make sense for you?

bastianccm:
@pedroigor I do understand, but from what I can tell the user profile configuration is not yet enabled by default, is that correct?
https://github.com/keycloak/keycloak/blob/main/common/src/main/java/org/keycloak/common/Profile.java#L67

Of course the setting does not work for quarkus either, but for everyone using up to Keycloak 16, the legacy setting does not have an effect anymore, from what I understand, fixing that was my intentation.

But feel free to close this if it's not making sense to update this part of the documentation anymore.

pedroigor:
> @pedroigor I do understand, but from what I can tell the user profile configuration is not yet enabled by default, is that correct? https://github.com/keycloak/keycloak/blob/main/common/src/main/java/org/keycloak/common/Profile.java#L67
> 

Yes, it is not. But the configuration can also be set to the declarative provider (the still preview feature).

> Of course the setting does not work for quarkus either, but for everyone using up to Keycloak 16, the legacy setting does not have an effect anymore, from what I understand, fixing that was my intentation.
> 

It works for the new dist too but uses a different format. The documentation is updated now to show this new format. 

We still have the same providers (legacy and declarative) in the new dist. And the legacy is still the default.

> But feel free to close this if it's not making sense to update this part of the documentation anymore.

Thanks, I appreciate your time on it.

bastianccm:
@pedroigor thanks for reviewing and clarifying!

PR #1336, opened 2021-12-17
Update release notes and migration guide for WildFly 25 upgrade

stianst:


stianst:
We will revert changes to https.adoc when we prepare branches for RH-SSO 7.6

vramik (commented):
LGTM, I'd suggest to add also note to https://www.keycloak.org/docs/latest/upgrading/index.html#_install_new_version ... I think a link to https://github.com/keycloak/keycloak-documentation/pull/1336/files#diff-e219212e71223c836862284e2ba0ae6b7e414a768037502e9c4ff166d5920834R5 will be sufficient. The reason for that is that if someone follows the guide from the beginning and therefore follows the steps he doesn't know anything about this issue. So something like `NOTE: For upgrage from Keycloak 15 or older please check Migrating to 16.0.0 section first`

vramik (approved):
Thank you @stianst LGTM

PR #1363, opened 2022-01-17
Add OIDC Back-channel logout endpoint

rjokelai:
I had trouble finding this in documentation when trying to set up two connected keycloak instances to perform a backchannel logout. I finally found the URL `/realms/{realm-name}/protocol/openid-connect/logout/backchannel-logout` by reading the source code of the LogoutEndpoint

https://github.com/keycloak/keycloak/blob/bfce612641a70e106b20b136431f0e4046b5c37f/services/src/main/java/org/keycloak/protocol/oidc/endpoints/LogoutEndpoint.java#L287

### DIFF
--- a/server_admin/topics/sso-protocols/con-server-oidc-uri-endpoints.adoc
+++ b/server_admin/topics/sso-protocols/con-server-oidc-uri-endpoints.adoc
@@ -17,6 +17,9 @@ https://localhost:8080/auth
 
 /realms/{realm-name}/protocol/openid-connect/logout::
   Used for performing logouts.
+  
+/realms/{realm-name}/protocol/openid-connect/logout/backchannel-logout::
+  Used for performing OIDC Back-channel logouts.
 
 /realms/{realm-name}/protocol/openid-connect/userinfo::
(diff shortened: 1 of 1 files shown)


andymunro (approved):

mposolda (changes requested):
I am sorry for late reply on this PR.

It should be somehow mentioned that this backchannel logout endpoint is supposed to be used only in the context of Identity brokering (in case that Keycloak itself is defacto used as an RP (client) to the identity provider.

With regards to this, it will be also good to document this endpoint in this section: https://www.keycloak.org/docs/latest/server_admin/index.html#identity-broker-logout and mention that this is useful just when external IDP want to trigger logout to the relying parties (clients) and that Keycloak was one of those clients.

Finally, it will be nice to add cross-links between both sections (this section and the IDP section), so people looking at one of the sections can look at the other section for more context.

ahus1:
The documentation has been moved to [the main Keycloak repository](https://github.com/keycloak/keycloak/tree/main/docs/documentation).

Please create an issue for this in the main repository, and then create a a new pull request there with these changes.

Thanks!

PR #1376, opened 2022-01-23
Update themes.adoc

psevestre:
It seems that the actual directory we have to create for a custom theme should start with "theme" instead of "themes".


andymunro (approved):

stianst:
Custom themes go into the `themes` directory in the distribution, so this is correct. The other PR that changed it to `theme` when the theme is wrapped-up as a JAR is correct. The inconsistency is not great, but it is what it is.

PR #1384, opened 2022-01-31
Add Quickstart project links for User Storage SPI

dusklight:
Adding links to the quickstart projects so that they can be easily found by readers.

Based it off from the Authorization Services document where it lists authorization quickstart projects.

It may not be obvious to the readers (such as myself) that there are official sample projects for implementing a User Storage SPI.  Having the links in the documentation would help to save time, as searching on the web brings up some old or incomplete examples.


andymunro (approved):

ahus1 (approved):
Thank you for these additions, they look good to me. I'll create a follow-up PR to use `{quickstartRepo_link}` for the two links.

mhajas (approved):
Approving per @andymunro 's and @ahus1 's approvals.

PR #1412, opened 2022-02-16
Extend documentation for password reset.

CollinAlpert:
I just spent a lot of time trying to figure out why Keycloak was throwing exceptions when trying to reset the password. Turns out, it was tripping over the "From" field being `null`:
image


ahus1:
Thanks for contributing this change and sorry that it took so long for us to review it. Could you please rebase your change as it is now conflicting with other changes that have already been merged?

PR #1422, opened 2022-02-25
Release and migration notes for 17.1 covering the topics db XA handli…

DGuhr:
…ng and JSON logging

### DIFF
--- a/release_notes/topics/17_1_0.adoc
+++ b/release_notes/topics/17_1_0.adoc
@@ -1,0 +1,20 @@
+= Highlights
+
+== title
+
+lorem ipsum...
+
+== Updates
+
+A lot of effort improving the Quarkus distribution to make it as good as an experience as possible. We want to thank all of you for your feedback, and we are working to make the experience even better in the future.
+
+A few highlights include:
+
+* Upgraded Quarkus to 2.7.2.Final
+* Better UX when using database vendors that do not support the XA transaction mode (e.g. Galera, Azure managed SQL). Now you
+only set `db=<db>` and `transaction-xa-enabled=false` to use a non-XA JDBC driver with non-XA transaction handling.
+* Initial support for JSON log output using `log-json-enabled` configuration property
+* ...
+
+== Other improvements
+dolor sit amet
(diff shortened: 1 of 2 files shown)


upgrading/topics/keycloak/changes-17_1_0.adoc
```diff
@@ -0,0 +1,18 @@
...
+* Container images for the legacy distribution tags have changed. To use the legacy distribution use the tags `legacy` or `18.0.0-legacy`.
+* Download on the website for the legacy distribution has changed to `keycloak-legacy-18.0.0.[zip|tar.gz]`.
+
+If you encounter problems migrating to the Quarkus distribution, missing ability to configure something, or have general ideas and feedback, please open a discussion in https://github.com/keycloak/keycloak/discussions/categories/keycloak-x-quarkus-distribution[GitHub Discussions].
+
+= Migrating from Keycloak 17.0.0
+
+A number of things have changed since Keycloak 17.0.0 was released. The changes include:
+
+* Improved UX for database vendors that don't support the XA transaction type (e.g. Galera, Azure managed MS-SQL): Setting the system properties `-Dkc.db-tx-type=enabled` and `-Dkc.db-driver=<nonxadriver>` don't work anymore. Instead, configure your database vendor and set `transaction-tx-enabled=false`.
```
stianst: Please see https://github.com/keycloak/keycloak-documentation/blob/main/upgrading/topics/keycloak/changes-17_0_0.adoc for reference on how to format this file.

It should be one header per change that is required.

release_notes/topics/17_1_0.adoc
```diff
@@ -0,0 +1,20 @@
...
+A few highlights include:
+
+* Upgraded Quarkus to 2.7.2.Final
+* Better UX when using database vendors that do not support the XA transaction mode (e.g. Galera, Azure managed SQL). Now you
+only set `db=<db>` and `transaction-xa-enabled=false` to use a non-XA JDBC driver with non-XA transaction handling.
+* Initial support for JSON log output using `log-json-enabled` configuration property
+* ...
+
+== Other improvements
+dolor sit amet
```
stianst, line 20: Please see previous files for reference on how to format this file. It's included in release blog as well as docs, so need to be formatted accordingly.

There should be a highlights, with subsections, and if relevant an other improvements. There must not be placeholders though as the file should be "ready to be released" always ;)

stianst (changes requested):

release_notes/topics/17_1_0.adoc
DGuhr: Just didn't think it's a highlight, so I didn't know what to write in the "highlights" section. Also not sure if we really want to promote it, as for me it is not clear galera/non-xa databases have first class support from our side. any idea? Should I just add the part around transactions in the highlights section instead?

PR #1440, opened 2022-03-21
Server Development updates - Forward port edits from Downstream MR

andymunro:
See https://issues.redhat.com/browse/CIAM-2488

Close #CIAM-2488

### DIFF
--- a/server_development/topics/admin-rest-api.adoc
+++ b/server_development/topics/admin-rest-api.adoc
@@ -2,25 +2,19 @@
 
 {project_name} comes with a fully functional Admin REST API with all features provided by the Admin Console.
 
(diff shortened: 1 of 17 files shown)


andymunro:
@nchopra  Can you approve these updates, which match the downstream updates? We need to keep the upstream version synch.

andymunro:
Errors are only for the Product version and only report that the build script cannot find the other guides. The log ends with a success message for all guides.

andymunro:
@mposolda Can you verify these changes to make this match the downstream version?

mposolda (approved):
@andymunro Thanks for the changes!

PR #1448, opened 2022-03-29
Update js adapter adoc to align w new proxy guide.

### LINKED ISSUE #10915: Documentation: conflict between best practices regarding the JavaScript adapter
### Describe the bug

The [JavaScript adapter's documentation](https://www.keycloak.org/docs/latest/securing_apps/#_javascript_adapter) recommends serving the adapter from the Keycloak server:

> A **best practice** is to load the JavaScript adapter **directly from Keycloak Server** as it will automatically be updated when you upgrade the server. If you copy the adapter to your web application instead, make sure you upgrade the adapter only after you have upgraded the server.

On the contrary, the Keycloak guide "Using a reverse proxy" recommends in its paragraph ["Exposed path recommendations"](https://www.keycloak.org/server/reverseproxy#_exposed_path_recommendations) to **not** expose the Keycloak `/js` path, and serve the JS adapter from the client:

> A **good practice** is to not use external js for the javascript client, but bake it into your public client instead.

This is contradictory. Of course, both options are possible. But Keycloak should recommend only one (with proper documentation) as a *best practice*, or just expose the pros/cons for each option.

### Version

17.0.0

### Expected behavior

N/A

### Actual behavior

N∕A

### How to Reproduce?

N/A

### Anything else?

N/A

DGuhr:
align with new guides recommendation. 

Closes keycloak/keycloak#10915

### DIFF
--- a/securing_apps/topics/oidc/javascript-adapter.adoc
+++ b/securing_apps/topics/oidc/javascript-adapter.adoc
@@ -3,11 +3,13 @@
 
 {project_name} comes with a client-side JavaScript library that can be used to secure HTML5/JavaScript applications. The JavaScript adapter has built-in support for Cordova applications.
 
-The library can be retrieved directly from the {project_name} server at `{kc_js_path}/keycloak.js` and is also distributed as a ZIP archive.
+A good practice is to include the JavaScript adapter in your application using a javascript package manager like npm or yarn. The keycloak-js package is available on the following locations:
 
-A best practice is to load the JavaScript adapter directly from {project_name} Server as it will automatically be updated when you upgrade the server. If you copy the adapter to your web application instead, make sure you upgrade the adapter only after you have upgraded the server.
+* NPM: https://www.npmjs.com/package/keycloak-js
+* Yarn: https://yarnpkg.com/package/keycloak-js
 
-NOTE: You can also download package from npm: https://www.npmjs.com/package/keycloak-js
+Alternatively, the library can be retrieved directly from the {project_name} server at `{kc_js_path}/keycloak.js` and is also distributed as a ZIP archive.
+
 
 One important thing to note about using client-side applications is that the client has to be a public client as there is no secure way to store client
 credentials in a client-side application. This makes it very important to make sure the redirect URIs you have configured for the client are correct and as specific as possible.
--- a/securing_apps/topics/oidc/javascript-adapter.adoc
+++ b/securing_apps/topics/oidc/javascript-adapter.adoc
@@ -3,7 +3,7 @@
 
 {project_name} comes with a client-side JavaScript library that can be used to secure HTML5/JavaScript applications. The JavaScript adapter has built-in support for Cordova applications.
 
-A good practice is to include the JavaScript adapter in your application using a javascript package manager like npm or yarn. The keycloak-js package is available on the following locations:
+A good practice is to include the JavaScript adapter in your application using a package manager like NPM or Yarn. The `keycloak-js` package is available on the following locations:
 
 * NPM: https://www.npmjs.com/package/keycloak-js
 * Yarn: https://yarnpkg.com/package/keycloak-js


securing_apps/topics/oidc/javascript-adapter.adoc
```diff
@@ -5,9 +5,9 @@
 
 The library can be retrieved directly from the {project_name} server at `{kc_js_path}/keycloak.js` and is also distributed as a ZIP archive.
 
-A best practice is to load the JavaScript adapter directly from {project_name} Server as it will automatically be updated when you upgrade the server. If you copy the adapter to your web application instead, make sure you upgrade the adapter only after you have upgraded the server.
+A good practice is to include the JavaScript adapter directly into your web application to make sure your web application is self-contained and starts up fast. Keep in mind you need to upgrade the JavaScript adapter after you have upgraded the server. Alternatively, load the JavaScript adapter directly from {project_name} Server. Then it will automatically be updated when you upgrade the server.
 
-NOTE: You can also download package from npm: https://www.npmjs.com/package/keycloak-js
+NOTE: You can also download the package from npm: https://www.npmjs.com/package/keycloak-js
```
jonkoops: I would personally start making this the recommended approach, as it allows us to introduce changes to the API without breaking the existing package.

securing_apps/topics/oidc/javascript-adapter.adoc
```diff
@@ -5,9 +5,9 @@
 
 The library can be retrieved directly from the {project_name} server at `{kc_js_path}/keycloak.js` and is also distributed as a ZIP archive.
 
-A best practice is to load the JavaScript adapter directly from {project_name} Server as it will automatically be updated when you upgrade the server. If you copy the adapter to your web application instead, make sure you upgrade the adapter only after you have upgraded the server.
+A good practice is to include the JavaScript adapter directly into your web application to make sure your web application is self-contained and starts up fast. Keep in mind you need to upgrade the JavaScript adapter after you have upgraded the server. Alternatively, load the JavaScript adapter directly from {project_name} Server. Then it will automatically be updated when you upgrade the server.
```
jonkoops: Copying the adapter code into your own application should never be a recommendation in my opinion. Folks should use a system that allows them to manage dependencies such as NPM so they can get timely updates and security warnings.

DGuhr: thx for your input! Makes sense for me. Then i'll update the text here and also create a PR to update https://www.keycloak.org/server/reverseproxy#_exposed_path_recommendations to state "A good practice is to not use external js for the javascript client, but add it into your public client instead. The keycloak js package is available at https://www.npmjs.com/package/keycloak-js" wdyt?

jonkoops: Yes, that sounds like a plan :+1:

DGuhr: @jonkoops I've put out an updated version, also linking the yarn package (didn't want to advertize one or the other, so added both ;) ) could you re-check?

securing_apps/topics/oidc/javascript-adapter.adoc
```diff
@@ -3,11 +3,13 @@
 
 {project_name} comes with a client-side JavaScript library that can be used to secure HTML5/JavaScript applications. The JavaScript adapter has built-in support for Cordova applications.
 
-The library can be retrieved directly from the {project_name} server at `{kc_js_path}/keycloak.js` and is also distributed as a ZIP archive.
+A good practice is to include the JavaScript adapter in your application using a javascript package manager like npm or yarn. The keycloak-js package is available on the following locations:
```
jonkoops:
```suggestion
A good practice is to include the JavaScript adapter in your application using a package manager like NPM or Yarn. The `keycloak-js` package is available on the following locations:
```

securing_apps/topics/oidc/javascript-adapter.adoc
```diff
@@ -3,11 +3,13 @@
...
 
-A best practice is to load the JavaScript adapter directly from {project_name} Server as it will automatically be updated when you upgrade the server. If you copy the adapter to your web application instead, make sure you upgrade the adapter only after you have upgraded the server.
+* NPM: https://www.npmjs.com/package/keycloak-js
+* Yarn: https://yarnpkg.com/package/keycloak-js
+
+Alternatively, the library can be retrieved directly from the {project_name} server at `{kc_js_path}/keycloak.js` and is also distributed as a ZIP archive.
```
jonkoops, line 11: It would be good to add a note that it is not recommended to load load the JavaScript from the server, as it might change at any point, possibly causing application code to misbehave.

jonkoops (dismissed):

securing_apps/topics/oidc/javascript-adapter.adoc
DGuhr: don't we only publish new js adapter versions together with new keycloak versions? So is the release cycle of the npm/yarn package tied to the keycloak server release cycle? If so, I'd say it is fine as is, because we say the preferred approach ("good practice") is using npm/yarn. If not, you are totally right and I'd add a warning.

DGuhr: .. but feel free to suggest something, i'll add :)

DGuhr:
@jonkoops here's the corresponding change to the proxy guide to make it more clear: https://github.com/keycloak/keycloak/pull/11006 would be nice if you could also approve, just one sentence.

securing_apps/topics/oidc/javascript-adapter.adoc
jonkoops: Yeah I'd say it's fine like this then. We do indeed tie to the Keycloak version with the NPM releases.

jonkoops (approved):

abstractj (approved):

PR #1449, opened 2022-03-29
Add images and new adoc files about client secret rotation

marcelomrwin:
https://github.com/keycloak/keycloak/issues/10610


andymunro:
@marcelomrwin 
One initiial question is if the focus should be on the new Admin Console because that will be the default console at Keycloak 19.  However, this feature may not be in the new console yet.  @ssilvert would know.  If we need both and it is on both consoles, these are the codes for the two consoles.

ifeval::["{kc_dist}" == "quarkus"]

endif::[]

ifeval::["{kc_dist}" == "wildfly"]

endif::[]

Actually, I believe we are not publishing a version of the Server Administrator Guide with WildFly-based content.  @stianst will know.

server_admin/topics/clients/oidc/con-secret-rotation.adoc
```diff
@@ -0,0 +1,28 @@
+[id="con-secret-rotation_{context}"]
+[[_secret_rotation]]
+= Client Secret Rotation
+[role="_abstract"]
+For client with <<_client-credentials, Confidential>> <<_access-type, Access Type>> {project_name} supports the functionality of rotating client secrets through <<_client_policies, Client Policies>>.
```
andymunro: For a client with <<_client-credentials, Confidential>> <<_access-type, Access Type>>, {project_name} supports the functionality of rotating client secrets through <<_client_policies, Client Policies>>.

server_admin/topics/clients/oidc/con-secret-rotation.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+The client secrets rotation policy provides greater security in order to alleviate problems such as secret leakage. Once enabled, {project_name} supports up to two concurrently active secrets for each client. The policy manages rotations according to the following settings:
+
+* *Secret expiration:* [seconds] - When the secret is rotated, this is the expiration of time of the new secret. The amount, _in seconds_, added to the secret creation date. Calculated at policy execution time.
+* *Rotated secret expiration:* [seconds] - When secret is rotated, this is the remaining expiration time for the old secret. This value should be always smaller than Secret expiration. When this is set to 0, the old secret will be immediately removed during client rotation. The amount, _in seconds_, added to the secret rotation date. Calculated at policy execution time.
```
andymunro: Add "the" before secret.
"this value is the remaining..."
"When the value is 0.."

server_admin/topics/clients/oidc/con-secret-rotation.adoc
```diff
@@ -0,0 +1,28 @@
+[id="con-secret-rotation_{context}"]
+[[_secret_rotation]]
+= Client Secret Rotation
+[role="_abstract"]
+For client with <<_client-credentials, Confidential>> <<_access-type, Access Type>> {project_name} supports the functionality of rotating client secrets through <<_client_policies, Client Policies>>.
+
+The client secrets rotation policy provides greater security in order to alleviate problems such as secret leakage. Once enabled, {project_name} supports up to two concurrently active secrets for each client. The policy manages rotations according to the following settings:
+
+* *Secret expiration:* [seconds] - When the secret is rotated, this is the expiration of time of the new secret. The amount, _in seconds_, added to the secret creation date. Calculated at policy execution time.
```
andymunro, line 26: Always follow "this" with a noun "this value..."

server_admin/topics/clients/oidc/con-secret-rotation.adoc
```diff
@@ -0,0 +1,28 @@
...
+* *Rotated secret expiration:* [seconds] - When secret is rotated, this is the remaining expiration time for the old secret. This value should be always smaller than Secret expiration. When this is set to 0, the old secret will be immediately removed during client rotation. The amount, _in seconds_, added to the secret rotation date. Calculated at policy execution time.
+* *Remaining expiration time for rotation during update:* [seconds] - Time period when an update to a dynamic client should perform client secret rotation. Calculated at policy execution time.
+
+Whenever there is a client secret rotation a new main secret is generated and the old client main secret becomes the secondary secret with a new expiration date.
```
andymunro: When a client secret rotation occurs, a new main secret

server_admin/topics/clients/oidc/con-secret-rotation.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+== Rules for client secret rotation
+Rotations do not occur automatically or through a background process. In order to perform the rotation, an update action is required on the client, either through the {project_name} Admin Console UI or Admin REST API. When invoking a client update action, secret rotation occurs according to the rules:
```
andymunro: Admin Console (without UI)

server_admin/topics/clients/oidc/con-secret-rotation.adoc
```diff
@@ -0,0 +1,28 @@
...
+* During dynamic client registration client-update request, the client secret will be automatically rotated if the value of  *Remaining expiration time for rotation during update* match the period between the current date and the *Secret expiration*.
+
+Additionally it is possible through Admin REST API to force a client secret rotation at any time.
+
+[NOTE]
+====
+During the creation of new clients, if the client secret rotation policy is active, the behavior will be applied automatically.
+====
+
+WARNING: To apply the secret rotation behavior in pre-existing clients, before creating the policy, it is necessary that after defining the policy, an update must be performed on the client so that the behavior is applied.
```
andymunro: WARNING: To apply the secret rotation behavior to an existing client, update that client after you define the policy so that the behavior is applied.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
+.Create a profile
+image:{project_images}/create-oidc-client-profile.png[Create Client Profile]
+
+. Enter any name for *Name*.
+
+. Enter a description that helps you identify the purpose of the profile for *Description*.
+
+. Click *Save*.
++
+This action creates the profile and enable to configure executors. Click *Create* to configure an executor for this profile.
```
andymunro: This action creates the profile and enables you to configure executors. 

. Click *Create* to configure an executor for this profile.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
++
+.Create a profile executors
+image:{project_images}/create-oidc-client-secret-rotation-executor.png[Client Profile Executor]
+
+. Select _secret-rotation_ for *Executor Type*
+
+. Enter the maximum duration time of each secret, in seconds, for *Secret Expiration*
```
andymunro: End with a period.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+. Enter the maximum duration time of each rotated secret, in seconds, for *Rotated Secret Expiration*.
++
+WARNING: Remember that *Rotated Secret Expiration* value must always be less than *Secret Expiration*
```
andymunro: Add "the" after "Remember that"
End with a period.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
+. Enter the amount of time, in seconds, after which any update action will update the client for *Remain Expiration Time*.
+
+. Click *Save*
```
andymunro: End with a period.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
++
+[NOTE]
+====
+In the example above:
+
+* Each secret is valid for one week
+* The rotated secret expires after two days 
+* The window for updating dynamic clients starts one day before the secret expires
+====
```
andymunro: End with a period because each item is a sentence.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
++
+. Return to the *Client Policies* tab
+
+. Click *Policies*
+
+. Click *Create*
++
```
andymunro: End with a period.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
+.Create the Client Secret Rotation Policy
+image:{project_images}/create-oidc-client-secret-rotation-policy.png[Client Rotation Policy]
+
+. Enter any name for *Name*.
+
+. Enter a description that helps you identify the purpose of the policy for *Description*.
+
+. Click *Save*.
++
+This action creates the policy and enable to associate policies with profiles. It also allows you to configure the conditions for policy execution.
```
andymunro: and enables you...

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
++
+. Under Conditions, click *Create*
```
andymunro: End with a period.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
++
+.Create the Client Secret Rotation Policy Condition
+image:{project_images}/create-oidc-client-secret-rotation-condition.png[Client Rotation Policy Condition]
+
+. To apply the behavior to all confidential clients select _client-access-type_ in the *Condition Type* field
++
+[NOTE]
+====
+To apply to a specific group of clients, another approach would be to select the _client-roles_ type in the *Condition Type* field. In this way, it would be possible to create specific roles and assign a custom rotation configuration to each role.
```
andymunro: Change "it would be possible to"
TO
"you could"

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
+====
++
+. Add _confidential_ to the field *Client Access Type*
```
andymunro: End with a period.

server_admin/topics/clients/oidc/proc-secret-rotation.adoc
```diff
@@ -0,0 +1,84 @@
...
+
+. Click *Save*
```
andymunro: End with a period.

andymunro (changes requested):
Looks good, @marcelomrwin.  I have a few changes to suggest.

server_admin/topics/clients/oidc/con-secret-rotation.adoc
mposolda: This is true in case that OIDC Dynamic client update request is used. But for admin REST API, the executor is triggered when new client is created or when client secret is re-generated (EG. Admin clicks on the "Re-generate secret" ). Is it correct? So maybe this part can be slightly re-formulated to mention that "Re-generate secret" is going to check client secret rotation instead of the client update in case of admin REST API?

server_admin/topics/clients/oidc/con-secret-rotation.adoc
mposolda: Similar to the comment above. I believe it can be "Re-generate secret" in case of admin console / admin REST API?

server_admin/topics/clients/oidc/con-secret-rotation.adoc
mposolda: Client secret rotation is preview feature and this fact probably worth to be mentioned in the beginning of this section. There is some snippet, which you can include to automatically address this. You can see for example "recovery-codes.adoc" for the inspiration when the similar snippet is used for the "Recovery codes" preview feature.

mposolda (changes requested):
@marcelomrwin Thanks for the docs Marcelo and thanks for adding the example! Looks nice to me.

I've added few comments inline. Besides that, there are few more places, which should be updated:
- profiles.adoc - The profile page of the "Server Installation guide" - the client secret rotation should be listed here as the preview feature
- Release notes ( release-notes/topics/18_0_0.adoc) : The client secret rotiation worth to be mentioned here in the release notes as preview feature similar to for example "Recovery codes" .

marcelomrwin:
> One initiial question is if the focus should be on the new Admin Console because that will be the default console at Keycloak 19. However, this feature may not be in the new console yet. @ssilvert would know. If we need both and it is on both consoles, these are the codes for the two consoles.
>
> ifeval::["{kc_dist}" == "quarkus"]

> endif::[]

> ifeval::["{kc_dist}" == "wildfly"]

> endif::[]

> Actually, I believe we are not publishing a version of the Server Administrator Guide with WildFly-based content. @stianst will know.

Hi @andymunro, All features were developed taking into account only the traditional UI. There is a team working to perform the migration to the new UI.

server_admin/topics/clients/oidc/con-secret-rotation.adoc
marcelomrwin: Sorry @mposolda, I don't think I quite understand what you mean. In this paragraph I would just like to mention that the rotation can be done either through the Admin Console or via REST. I didn't think the detail of the executioner's performance made any difference.

Maybe mention that the "Re-generate secret" button performs an update on the client?

server_admin/topics/clients/oidc/con-secret-rotation.adoc
marcelomrwin: You say, simplify the explanation and use the action of "Re-generate secret" by REST or Admin UI?

server_admin/topics/clients/oidc/con-secret-rotation.adoc
marcelomrwin: thanks @mposolda

server_admin/topics/clients/oidc/con-secret-rotation.adoc
mposolda: @marcelomrwin Yes, exactly. That is what I meant.

Please correct me if I am wrong here: I assume example scenario when administrator did setup of the client policy with the executor and he wants to apply it on the existing clients through admin REST API / admin console. Then the administrator needs to:
- Go to some client in the admin console
- Go to tab "Credentials"
- Click "Re-generate secret" .
Am I correct here in the description of this procedure?

If yes, I suggest to mention that "Re-generate secret" is needed instead of just "Updating client" .

Updating client implies to me that administrator needs to do some change in the initial screen of the client (Tab "Details" of the client in the admin console) and click "Save" . But AFAIK updating client this way in the admin console is not sufficient to re-generate and rotate secret. Or am I wrong here?

server_admin/topics/clients/oidc/con-secret-rotation.adoc
mposolda: > Maybe mention that the "Re-generate secret" button performs an update on the client?

Yes, exactly. That will be probably good. As for the admin console, the "update of the client" implies to me that I just click on the client, update some settings of the client in the tab "Details" and click "Save" . Which won't help in this case if I understand correctly?

marcelomrwin: @mposolda You are right.

marcelomrwin: Let me explain better on the comment below.

server_admin/topics/clients/oidc/con-secret-rotation.adoc
marcelomrwin: I chose to insert this information in the server_admin/topics/clients/oidc/proc-secret-rotation.adoc file because how it reports a process or a flow makes more sense.

server_admin/topics/clients/oidc/con-secret-rotation.adoc
marcelomrwin: Updated

mposolda:
@marcelomrwin Thanks for the update. I have 2 points:
- Release notes is still missing. You can update them in this file: https://github.com/keycloak/keycloak-documentation/blob/main/release_notes/topics/18_0_0.adoc . TIP: I suggest to add them at the beginning of this file to avoid conflicts with Doug's PR :-) That one is here fyi: https://github.com/keycloak/keycloak-documentation/pull/1467 

- I see some test failures and looks they are regression of this PR. I guess that you can just move the image files to this directory "images" instead of "keycloak-images" : https://github.com/keycloak/keycloak-documentation/tree/main/server_admin/images . The directory "keycloak-images" is only for images, which contains keycloak logo in the header. They need to have special / different version in the "rhsso-images" where they need to have RH-SSO logo.
Since your images don't have logo, the "images" directory should be fine.

release_notes/topics/18_0_0.adoc
```diff
@@ -31,6 +31,12 @@ Thanks to https://github.com/vanrar68[Joaquim Fellmann] for the contribution.
 
 There are more WebAuthn improvements and fixes in addition to that.
 
+== Client secret rotation
+
+{project_name} now supports Client Secret Rotation through customer policies. This feature allows thats confidential clients can be provided with realm policies allowing the use up to two secrets simultaneously.
```
mposolda: Is it possible to add note that this is preview feature? See "Recovery Codes" section for the inspiration :-)

marcelomrwin: Done

mposolda (approved):
@marcelomrwin Thanks for the fix!

Issue #1494, opened 2022-04-20
Account Console chapter


andymunro:
### Description

Update this Server Administration Guide chapter for Account Console changes

agagancarczyk:
hi @andymunro I can take a look at this issue, but could you please confirm what chapter needs to be updated? Is it https://www.keycloak.org/docs/latest/server_admin/#keycloak-features-and-concepts ?Thanks

andymunro:
Sure, @agagancarczyk.  We need to check if the Account Console is up-to-date, which is actually an independent exercise of updating the Server Administration Guide for the new Admin Console.
Anyway, here is the chapter with that content; https://raw.githubusercontent.com/keycloak/keycloak-documentation/main/server_admin/topics/account.adoc

PR #1546, opened 2022-05-31
Updated account console chapter #1494

agagancarczyk:
This PR is for https://github.com/keycloak/keycloak-documentation/issues/1494.

Updated all sections and screenshots except for:
==== Two-factor authentication with WebAuthn 
==== Passwordless authentication with WebAuthn


andymunro (changes requested):
Looks great, @agagancarczyk. But is there any way to reduce the size of the screens before capturing them? It would be great if the screens lunch could be smaller and the text on the screens is closer to the size of the text in the instructions.  
If you cannot resize them before capturing them, I might be able to use GIMP to shrink them.

agagancarczyk:
> Looks great, @agagancarczyk. But is there any way to reduce the size of the screens before capturing them? It would be great if the screens lunch could be smaller and the text on the screens is closer to the size of the text in the instructions. If you cannot resize them before capturing them, I might be able to use GIMP to shrink them.

hi @andymunro I have resized all new screenshots. Hope they look good now.

andymunro (changes requested):
Thanks, @agagancarczyk. Looks great. However, needs to be merged to upstream/new-admin-console.

andymunro (approved):
Actually, on second thought, this PR is not really related to the new Admin Console, so why don't we just merge this to main. 
Okay to merge @ssilvert

ssilvert (approved):

edewit:
@ssilvert this one also went into main instead of `new-admin-console`

andymunro:
Yes, thanks, @edewit.  @ssilvert and I decided it's okay, because it is actually a valid update independent of the new Admin Console.

PR #1548, opened 2022-06-01
Issue #9194: Client authentication fails when using signed JWT, if the JWA signing algorithm is not RS256

rmartinc:
Issue: https://github.com/keycloak/keycloak/issues/9194

Just a little modification to add the `algorithm` field for the `jwt` credentials configuration for java clients. You can change any word you like or ask me to do it.

Thanks!

### DIFF
--- a/securing_apps/topics/oidc/java/client-authentication.adoc
+++ b/securing_apps/topics/oidc/java/client-authentication.adoc
@@ -45,6 +45,7 @@ For set up on the adapter side you need to have something like this in your `key
     "client-keystore-password": "storepass",
     "client-key-password": "keypass",
     "client-key-alias": "clientkey",
+    "algorithm": "RS256",
     "token-expiration": 10
   }
 }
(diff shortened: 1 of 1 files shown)


securing_apps/topics/oidc/java/client-authentication.adoc
```diff
@@ -53,6 +54,10 @@ For set up on the adapter side you need to have something like this in your `key
 With this configuration, the keystore file `keystore-client.jks` must be available on classpath in your WAR. If you do not use the prefix `classpath:`
 you can point to any file on the file system where the client application is running.
 
+The `algorithm` field specifies the algorithm used for the Signed JWT and it defaults to `RS256`. This field should be in sync with the key pair.
+For example algorithm RS256 needs RSA keys but ES256 requires EC keys. Please refer to
```
tnorimat: How about the following?

For example, the `RS256` algorithm needs a RSA key pair while the `ES256` algorithm requires an EC key pair

tnorimat (commented):
@rmartinc I added one review comment. Could you check it?

securing_apps/topics/oidc/java/client-authentication.adoc
rmartinc: It's OK to me! Modified!

mposolda:
@rmartinc @tnorimat Thanks for the documentation update and for the review!

PR #1581, opened 2022-06-21
Update token exchange documentation with latest changes to the audience constraints

### LINKED ISSUE #12634: Update token exchange documentation with latest changes to the audience constraints
### Description

Token exchange is now more strict about the audience and the documentation is not reflecting the changes recently made.

### Discussion

https://github.com/keycloak/keycloak/issues/12438

### Motivation

https://github.com/keycloak/keycloak/issues/12438

### Details

https://github.com/keycloak/keycloak/issues/12438

pedroigor:
Closes [#12634](https://github.com/keycloak/keycloak/issues/12634)

### DIFF
--- a/securing_apps/topics/token-exchange/token-exchange.adoc
+++ b/securing_apps/topics/token-exchange/token-exchange.adoc
@@ -123,7 +123,6 @@ The rest of this chapter discusses the setup requirements and provides examples 
 For simplicity's sake, let's call a token minted by the current realm as an _internal_ token and a token minted by
 an external realm or identity provider as an _external_ token.
 
-
 === Internal token to internal token exchange
 
 With an internal token to token exchange you have an existing token minted to a specific client and you want to exchange
@@ -196,10 +195,18 @@ curl -X POST \
     http://localhost:8080{kc_realms_path}/myrealm/protocol/openid-connect/token
 ----
 
-
 The `subject_token` parameter must be an access token for the target realm.  If your `requested_token_type` parameter
 is a refresh token type, then the response will contain both an access token, refresh token, and expiration.  Here's
 an example JSON response you get back from this call.
+
+When the `audience` parameter is not set, the value of the parameter defaults to the client making the token exchange request.
+
+Unlike with confidential clients, public clients are not allowed to perform token exchanges using tokens from other clients.
+If you are passing a `subject_token`, the (confidential) client that was issued the token should either match the client making the request or, if issued to a different client,
+the client making the request should be among the audiences set to the token.
+
+If you are explicitly setting a target `audience` (with a client different from the client making the request), you should also make sure that the `token-exchange` scope permission is configured for the client set to the `audience` parameter to allow
+the client making the request to successfully complete the exchange.
 
 [source,json]
 ----


securing_apps/topics/token-exchange/token-exchange.adoc
```diff
@@ -194,11 +193,18 @@ curl -X POST \
...
 
-
 The `subject_token` parameter must be an access token for the target realm.  If your `requested_token_type` parameter
 is a refresh token type, then the response will contain both an access token, refresh token, and expiration.  Here's
 an example JSON response you get back from this call.
 
+When the `audience` parameter is not set, the value of the parameter defaults to the client making the token exchange request. If you are passing a `subject_token`, the client to which the token was issued for should either match the client making the request or, if issued to a different client,
+the client making the request should be among the audiences set to the token.
+
+If you are explicitly setting a target `audience`, you should also make sure that the `token-exchange` scope permission is configured for the client set to the `audience` parameter to allow
```
dreamcrash: Does this imply that I can make an exchange between two clients where none of them have the permissions to perform the token exchange, as long as I don't explicitly set the audience?

securing_apps/topics/token-exchange/token-exchange.adoc
```diff
@@ -194,11 +193,18 @@ curl -X POST \
...
+the client making the request to successfully complete the exchange.
+
+Public clients are not allowed to perform token exchanges using tokens from other clients.
```
dreamcrash: In line 200 you have "if issued to a different client", now looking at the line 206, this means that in line 200 that phrase only applies to confidential clients. Maybe it would be more clear moving the content of line 206 to line 200.
Meaning starting immediately with "Public clients are not allowed to perform token exchanges using tokens from other clients."

securing_apps/topics/token-exchange/token-exchange.adoc
pedroigor: Not really, if you don't set an audience it is like the client is exchanging a token to itself. The previous statement tries to cover this.

Please, let me know if it needs more clarification.

securing_apps/topics/token-exchange/token-exchange.adoc
pedroigor: Makes sense.

pedroigor: @dreamcrash I think I did it right, please let me know what you think. Basically, we first state that public clients are not allowed to then describe the usage/constraints around `subject_token.

dreamcrash: It looks good, but I am wondering if we can improved a bit more. The first time I read the token-exchange documentation I had a hard time understanding all the cases for the internal-to-internal token exchange because there are several variables into play, namely: public vs confidential; token holder vs non-token holder; target audience vs non target audience; in the token audience list vs not in the token audience  list.

I have compiled the following diagrams (I hope I did not made a mistake) based on what I have learn by looking into the code and doing my own testing:

Public clients: https://ibb.co/YN2rnr2
Confidential clients: https://ibb.co/1f15mzy

When I had a look the first time at your PR I was not fully aware of all the use-cases, now I think I can give you a better review.

My suggestions in bold:

**Unlike with confidential clients,** public clients are not allowed to perform token exchanges using tokens from other clients. --> The first part makes it a bit more explicit that confidential clients can perform token exchanges even if they are not the token holders.

If you are passing a `subject_token`, the client to which the token was issued for should either match the client making the request or, if issued to a different client, the **(confidential)** client making the request should be among the audiences set to the token. --> Just to reenforce again that the second case only applies for confidential clients.

If you are explicitly setting a target `audience` **(with a client different than the client making the request)**, you should also make sure that the `token-exchange` scope permission is configured for the client set to the `audience` parameter to allow the client making the request to successfully complete the exchange. --> Here I am being a bit pedantic but it is true that if audience=starting client then we do not need any permission.

If you feel that the original text is good enough feel free to ignore my suggestions.

dreamcrash (dismissed):

securing_apps/topics/token-exchange/token-exchange.adoc
pedroigor: They make sense and I appreciate the diagrams and your time on this. Changes done.

dreamcrash (dismissed):

securing_apps/topics/token-exchange/token-exchange.adoc
```diff
@@ -211,11 +210,19 @@ curl -X POST \
...
 
-
 The `subject_token` parameter must be an access token for the target realm.  If your `requested_token_type` parameter
 is a refresh token type, then the response will contain both an access token, refresh token, and expiration.  Here's
 an example JSON response you get back from this call.
 
+When the `audience` parameter is not set, the value of the parameter defaults to the client making the token exchange request.
+
+Unlike with confidential clients, public clients are not allowed to perform token exchanges using tokens from other clients.
+If you are passing a `subject_token`, the (confidential) client to which the token was issued for should either match the client making the request or, if issued to a different client,
```
andymunro:
```suggestion
If you are passing a `subject_token`, the (confidential) client that was issued the token should either match the client making the request or, if issued to a different client,
```

securing_apps/topics/token-exchange/token-exchange.adoc
```diff
@@ -211,11 +210,19 @@ curl -X POST \
...
+the client making the request should be among the audiences set to the token.
+
+If you are explicitly setting a target `audience` (with a client different than the client making the request), you should also make sure that the `token-exchange` scope permission is configured for the client set to the `audience` parameter to allow
```
andymunro:
```suggestion
If you are explicitly setting a target `audience` (with a client different from the client making the request), you should also make sure that the `token-exchange` scope permission is configured for the client set to the `audience` parameter to allow
```

andymunro (changes requested):
Just two minor changes, @pedroigor 
thanks

pedroigor:
@andymunro Thanks. Changes applied.

andymunro (approved):
Looks good, @pedroigor. thanks.

andymunro (approved):

dreamcrash (dismissed):

andymunro (dismissed):
Looks good, @pedroigor

andymunro (approved):
Looks good, @pedroigor

vmuzikar (approved):
Approving based on @andymunro's review.

PR #1586, opened 2022-06-24
Fix duplicate sentences and grammar in LDAP user federation doc

### LINKED ISSUE #1585: LDAP user federation troubleshooting documentation issue
### Describe the bug

LDAP user federation troubleshooting documentation includes duplicate sentences:

When you create LDAP provider, message appear in the server log in the INFO level starting with:
When you create LDAP provider, message appear in the server log in the INFO level starting with:

For tracking the performance or connection pooling issues, consider setting the value of property Connection Pool Debug Level of
For tracking the performance or connection pooling issues, consider setting the value of property Connection Pool Debug Level of

For the case of reporting LDAP issue, you may consider to attach some part of your LDAP tree with the target data, which causes issues in your environment. For example if login of some user takes lot of time, you can consider attach his LDAP entry showing count of member attributes of various "group" entries. In this case, it might be useful to add if those group entries are mapped to some Group LDAP mapper (or Role LDAP Mapper) in {project_name} etc.
For the case of reporting LDAP issue, you may consider to attach some part of your LDAP tree with the target data, which causes issues in your environment. For example if login of some user takes lot of time, you can consider attach his LDAP entry showing count of member attributes of various "group" entries. In this case, it might be useful to add if those group entries are mapped to some Group LDAP mapper (or Role LDAP Mapper) in {project_name} and so on.

### Version

18.0.1

### Expected behavior

Only one occurrence of the sentence should present

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

I'll submit a PR to fix the duplicates identified above

vanrar68:
Closes #1585

### DIFF
--- a/server_admin/topics/user-federation/ldap.adoc
+++ b/server_admin/topics/user-federation/ldap.adoc
@@ -156,9 +156,8 @@ enabled TRACE logging. If it is too big, the good alternative is to include just
 added to the log during the operation, which causes the issues to you.
 
 
-- When you create LDAP provider, message appear in the server log in the INFO level starting with:
+- When you create an LDAP provider, a message appears in the server log in the INFO level starting with:
 
- When you create LDAP provider, message appear in the server log in the INFO level starting with:
 
 ```
 Creating new LDAP Store for the LDAP storage provider: ...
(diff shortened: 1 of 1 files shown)


ahus1 (approved):
LGTM, thank you for taking the time to fix this!

andymunro (approved):

hmlnarik (approved):

PR #1607, opened 2022-07-04
Refactor UserStorageUtil usage.

ahus1:
The paragraph describing the old way is now no longer necessary and has been removed. The new way apply now also to the two methods mentioned earlier.

Relates to keycloak/keycloak#12805


mhajas (approved):
Thanks @ahus1, LGTM

abstractj (approved):

PR #1619, opened 2022-07-17
Add detail example for adding 'User Session Count Limit' execution to the browser authentication flow

HwangTaehyun:
Hi, Thank you for the amazing open source.

Recently, I added user session limit execution flow with my keycloak setting with the following reference.

* https://www.keycloak.org/docs/latest/server_admin/index.html#_step-up-flow:~:text=throw%20an%20error.-,User%20session%20limits,-Limits%20on%20the
* https://github.com/keycloak/keycloak/issues/10077

However, It isn't very clear for me to add this feature to the authorization flow.
It would be more helpful if there are more examples or additional explanations.

image

Thank you! @mposolda @andymunro

Please review this PR @mposolda @andymunro

### DIFF
--- a/server_admin/topics/authentication/flows.adoc
+++ b/server_admin/topics/authentication/flows.adoc
@@ -19,7 +19,7 @@ image:{project_images}/browser-flow.png[Browser Flow]
 Two sections exist.
 
 ===== Auth type
-The name of the authentication or the action to execute. If an authentication is indented, it is in a sub-flow. It may or may not be executed, depending on the behavior of its parent. 
+The name of the authentication or the action to execute. If an authentication is indented, it is in a sub-flow. It may or may not be executed, depending on the behavior of its parent.
 
 . Cookie
 +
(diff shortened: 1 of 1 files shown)


server_admin/topics/authentication/flows.adoc
```diff
@@ -424,7 +424,7 @@ To add session limits to a flow, perform the following steps.
 
 . Optionally, add a custom error message to be displayed when the limit is reached.
 
-Note that the user session limits should be added to your bound *Browser flow*, *Direct grant flow*, *Reset credentials* and also to any *Post broker login flow*.
+Note that the user session limits execution should be added to your bound *Browser flow*, *Direct grant flow*, *Reset Credentials* and also to any *Post broker login flow*. If you want to add the user session limits execution to *Browser Flow*, you should add this execution to *Forms Auth Type* by using *Actions* button.
```
andymunro:
```suggestion
Note that the user session limits execution should be added to your bound *Browser flow*, *Direct grant flow*, *Reset Credentials* and also to any *Post broker login flow*. To add the user session limits execution to *Browser Flow*, you should add this execution to *Forms Auth Type* by using the *Actions* button.
```

andymunro (changes requested):
@HwangTaehyun One minor change to suggest.

HwangTaehyun:
@andymunro Thanks for the review! I've applied your suggestions. Also, I integrated the detailed explanation to add user session limits execution with #1662

mposolda:
@HwangTaehyun Thanks for the PR. I am closing it as the section of `User Session Limits` was updated in the meantime and it should be clearer how to configure "User Session Limits" . BTV. For the proper behaviour, it may be needed to add `User Session Limits` as the REQUIRED authenticator at the level 1 as described in the current docs and the example flow (The case you described when adding to the `Forms` flow won't work for some cases like for example authentication through Kerberos or X509)

PR #1623, opened 2022-07-21
Promote new admin console to default

stianst:


mposolda (dismissed):
@stianst This looks good to me. Few minor points, but leaving to you to doublecheck if they are fine with this PR:
- Regarding migration, I guess that if people have explicitly set `keycloak` as the admin console theme in their realms (or their own theme, which is child of`keycloak` theme), they will probably still see the old admin console. Not sure if to somehow mention this fact as well in the migration notes?

- Docs still contains lots of screenshots with the old admin console. Do we want to replace them? But at the same time, it looks like fair amount of work and maybe not so much added value in it as most of screenshots are likely almost same/similar in the new admin console?

stianst:
> @stianst This looks good to me. Few minor points, but leaving to you to doublecheck if they are fine with this PR:
> 
> * Regarding migration, I guess that if people have explicitly set `keycloak` as the admin console theme in their realms (or their own theme, which is child of`keycloak` theme), they will probably still see the old admin console. Not sure if to somehow mention this fact as well in the migration notes?

Yeah, I agree, will add something around this.

> * Docs still contains lots of screenshots with the old admin console. Do we want to replace them? But at the same time, it looks like fair amount of work and maybe not so much added value in it as most of screenshots are likely almost same/similar in the new admin console?

Good point, honestly completely forgot about this part. From what I know this is a work in progress, but not sure what the status is. @ssilvert @andymunro where're we at with updating docs to the new admin console?

ssilvert:
@stianst  I think we are good to go with updated docs.  The team has been working on this for quite awhile.  But I haven't heard confirmation from @andymunro if that last PR I merged this morning is truly the end.

PR #1662, opened 2022-08-11
Improve documentation for user session limits

### LINKED ISSUE #1661: Documentation for user-session-limits is in incorrect section and should be improved
### Description

Documentation for `user-session-limits` is incorrectly wrapped into step-up section. Also this documentation is not very clear.

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mposolda:
Closes #1661


server_admin/topics/authentication/flows.adoc
```diff
@@ -415,16 +415,26 @@ To add session limits to a flow, perform the following steps.
...
 . Click *⚙️* (gear icon) for the *User Session Count Limiter*.
 . Enter an alias for this config.
-. Enter the required maximum number of sessions a user can have in this realm. If a value of 0 is used, this check is disabled.
-. Enter the required maximum number of sessions a user can have for the client. If a value of 0 is used, this check is disabled.
+. Enter the required maximum number of sessions a user can have in this realm. For example if configured to 2, then each user can have at
+maximum 2 SSO sessions in this realm. If a value of 0 is used, this check is disabled.
+. Enter the required maximum number of sessions a user can have for the client. For example if configured to 2, then each user can have at
+maximum 2 SSO sessions in this realm for each client. So when user is trying to authenticate to client `foo`, but he already has 2 SSO sessions where he is authenticated
+to client `foo`, then the authentication won't be permitted (or one of the existing sessions will be killed based on the behaviour configured). If a value of 0 is used, this check is disabled.
```
andymunro:
```suggestion
. Enter the required maximum number of sessions that a user can have in this realm. For example, if 2 is the value, 2 SSO sessions is the maximum that each user can have in this realm. If 0 is the value, this check is disabled.
. Enter the required maximum number of sessions a user can have for the client. For example, if 2 is the value, then 2 SSO sessions is the maximum in this realm for each client. So when a user is trying to authenticate to client `foo`, but that user has already authenticated in 2 SSO sessions to client `foo`, either the authentication will be denied or an existing sessions will be killed based on the behavior configured. If a value of 0 is used, this check is disabled.
```

server_admin/topics/authentication/flows.adoc
```diff
@@ -415,16 +415,26 @@ To add session limits to a flow, perform the following steps.
...
+If both session limits and client session limits are enabled, it makes sense to have client session limits to be always lower than session limits as limit per client can never be bigger than limit of all SSO sessions of this user.
```
andymunro:
```suggestion
If both session limits and client session limits are enabled, it makes sense to have client session limits to be always lower than session limits. The limit per client can never exceed the limit of all SSO sessions of this user.
```

server_admin/topics/authentication/flows.adoc
```diff
@@ -415,16 +415,26 @@ To add session limits to a flow, perform the following steps.
...
 
-  Deny new session - when a new session is requested and the session limit is reached, no new sessions can be created.
-  Terminate oldest session - when a new session is requested and the session limit has been reached, the oldest session will be removed and the new session created.
+  - *Deny new session* - when a new session is requested and the session limit is reached, no new sessions can be created.
+  - *Terminate oldest session* - when a new session is requested and the session limit has been reached, the oldest session will be removed and the new session created.
 
 . Optionally, add a custom error message to be displayed when the limit is reached.
 
 Note that the user session limits should be added to your bound *Browser flow*, *Direct grant flow*, *Reset credentials* and also to any *Post broker login flow*.
+Authenticator should be added at the point when the user is already known during authentication (usually at the end of the authentication flow) and should be typically REQUIRED. Note that it is not possible to have
```
andymunro:
```suggestion
The authenticator should be added at the point when the user is already known during authentication (usually at the end of the authentication flow) and should be typically REQUIRED. Note that it is not possible to have
```

server_admin/topics/authentication/flows.adoc
```diff
@@ -415,16 +415,26 @@ To add session limits to a flow, perform the following steps.
...
+ALTERNATIVE and REQUIRED executions at the same level and hence for example for the default browser flow, it may be needed to wrap existing flow as a REQUIRED level-1 subflow and then
+add `User Session Count Limiter` to the same level as this new subflow. Example of such flow is below.
+
```
andymunro:
```suggestion
ALTERNATIVE and REQUIRED executions at the same level. For example for the default browser flow, it may be necessary to wrap the existing flow as a REQUIRED level-1 subflow and 
add `User Session Count Limiter` to the same level as this new subflow. Example of such flow is below.

```

andymunro (changes requested):
A few suggestions, @mposolda

mposolda:
@andymunro Thanks a lot for the review! I've applied all your suggestions.

PR #1663, opened 2022-08-12
Update auth-spi.adoc

KirillAmber:
Fix the source of ftl

### DIFF
--- a/server_development/topics/auth-spi.adoc
+++ b/server_development/topics/auth-spi.adoc
@@ -809,7 +809,7 @@ The value of an attribute can be any Java bean as well.
 
 If you look at the top of the file, you'll see that we are importing a template:
 
-[source,ftl]
+[source,xml]
 ----
 <#import "select.ftl" as layout>
 ----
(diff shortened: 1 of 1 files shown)


stianst (dismissed):

stianst (commented):
Don't think ftl is a known highlighter, so maybe need to use XML instead?

server_development/topics/auth-spi.adoc
```diff
@@ -776,7 +776,7 @@ This file should be added to the `theme-resources/templates` in your JAR, see <<
 
 Let's take a bigger look at secret-question.ftl  Here's a small code snippet:
 
-[source,java]
+[source,ftl]
```
stianst:
```suggestion
[source,xml]
```

server_development/topics/auth-spi.adoc
```diff
@@ -809,7 +809,7 @@ The value of an attribute can be any Java bean as well.
 
 If you look at the top of the file, you'll see that we are importing a template:
 
-[source,java]
+[source,ftl]
```
stianst:
```suggestion
[source,xml]
```

PR #1664, opened 2022-08-12
Add info about consent when revoking refresh tokens

derkoe:
See also https://github.com/keycloak/keycloak/issues/12916


mposolda (approved):
@derkoe Thanks!

PR #1673, opened 2022-09-01
View groups from account console

cgeorgilakis:
Closes #8748

[Related Keycloak PR](https://github.com/keycloak/keycloak/pull/7933)


server_admin/topics/account.adoc
```diff
@@ -147,3 +147,15 @@ The *Applications* menu item shows users which applications you can access. In t
...
 image:{project_images}/account-console-applications.png[Applications]
+
+=== Viewing group memberships
+
+You can view the groups you are associated with by clicking the *Groups* menu.
+If you select *Direct membership* checkbox, you will see only the groups you are direct associated with.
+
+.Prerequisites
+
+* You need to have the *view-groups* account role for being able to view *Groups* menu.
```
stianst, line 158: I also found that I had to add scope on account/view-groups role to the account-console client which is used by account2. I think that's fine as long as we document it here though.

cgeorgilakis: Based on changes in Keycloak PR, I believe here we do not need any change.

stianst (approved):

PR #1718, opened 2022-11-14
Fix formatting of ${vault.key} variables

roman-mueller:
Since the `${vault.key}` is part of a `source` block, it is currently rendered confusingly in the documentation. See: https://www.keycloak.org/docs/latest/server_admin/#_vault-administration

I think removing all markup from it is the correct way to display it there.

### DIFF
--- a/server_admin/topics/vault.adoc
+++ b/server_admin/topics/vault.adoc
@@ -7,7 +7,7 @@ To obtain a secret from a vault rather than entering it directly, enter the foll
 
 [source]
 ----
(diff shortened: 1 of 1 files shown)


andymunro (approved):

ahus1:
Thanks for creating this PR and sorry for the late reply. In the meantime, this has been fixed by commit 1e380ba0f52c5a3e12fd06faf960837dbd082048.

Thanks again for reporting this.

PR #1720, opened 2022-11-21
Document how to force refresh access token in JS adapter

PinPinIre:
I was reading the documentation of the Javascript adapter to try and work out how to forcibly refresh an access token. There was no mention in the documentation on how to do this but I noticed in the actual code that this can be achieved by passing -1 to the updateToken function (https://github.com/keycloak/keycloak-ui/blob/main/libs/keycloak-js/src/keycloak.js#L615).

This PR updates the documentation to make it clearer how an access token can be forcibly refreshed.


andymunro (approved):

ahus1 (dismissed):
Thank you for adding this to the documentation!

andymunro (dismissed):

PR #1734, opened 2022-12-22
remove formatting characters from vault key literal code

carlin-q-scott:
The renderer for the hosted documentation was not handling the markdown formatting characters as expected, resulting in an incorrect example for the key reference.

The asterisks and underscores should not have been shown in the key reference example. I removed the underscores from the other key references because I think they were confusing when reading the markdown and I thought it was unnecessary to italicize them.

image

### DIFF
--- a/server_admin/topics/vault.adoc
+++ b/server_admin/topics/vault.adoc
@@ -7,11 +7,11 @@ To obtain a secret from a vault rather than entering it directly, enter the foll
 
 [source]
 ----
-**${vault.**_key_**}**
+${vault.key}
 ----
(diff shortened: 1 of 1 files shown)


andymunro (approved):

ahus1 (approved):
Thanks for fixing this!

As an alternative for the future, AsciiDoc allows formatting in source code snippets via `[source,subs="+quotes"]`.

Still, this PR is good as it is.

hmlnarik (approved):
Approving per @andymunro 's approval

PR #1753, opened 2023-01-26
Metrics enabled for Infinispan

ahus1:
keycloak/keycloak#15901

@sschu - this adds a release notes entry for KC21 that metrics for Infinispan are now available via the metrics endpoint. 

I also want to declare them as a preview feature, as those metrics names can change (and probably will change looking at their names). At the same time, those changes are probably due in the upstream project, like hiding the node name as a dimension and removing the duplicated cache name from the metric.

### DIFF
--- a/release_notes/topics/21_0_0.adoc
+++ b/release_notes/topics/21_0_0.adoc
@@ -11,3 +11,8 @@ See the migration guide for details.
 We removed the out-of-box support for Hashicorp vault in this release.
 
 See this https://github.com/keycloak/keycloak/discussions/16446[discussion] for more details.
+
+= Metrics for embedded Infinispan as preview
+
+When metrics are enabled and {project_name} uses the embedded {jdgserver_name}, the metrics endpoint exposes information about caches using the `/metrics` endpoint in the metrics with names starting with `vendor_cache_*`.
+Those metrics are currently a preview feature and might change in upcoming releases.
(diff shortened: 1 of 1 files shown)


sschu:
@ahus1 Thanks, LGTM!

vmuzikar (dismissed):
@ahus1 Thank you, makes sense to me to mark it as a preview feature, even more if we plan to decouple Infinispan from Keycloak in the future. This leaves us the room for changes,

pedroigor (commented):
I don`t think we should mark it as a preview but mention the format for the individual metrics is based on the original ISPN format.

It is not different than what we do in the cache configuration, which is also based on original ISPN configuration schema and it is also subject to change.

ahus1:
@pedroigor - I slightly update the docs with your comment, and asking you to re-review the added paragraph.

I'd like to see both of our statements in the docs: That it is preview and that it will change with Infinispan releases. 

The primary reason for preview that I see: Those metrics are not very Prometheus style at the moment and I want to actively push for more Prometheus-style metrics.

ahus1:
I'm closing this PR as I found that metrics could be enabled in KC20 the same way it is now planned for in KC21 and after https://github.com/keycloak/keycloak/pull/16630 has been merged. 

Part of the PR are some additional docs, and I don't think this alone doesn't make the case for adding something to the release notes.

ahus1:
I'm closing this PR as I found that metrics could be enabled in KC20 the same way it is now planned for in KC21 and after https://github.com/keycloak/keycloak/pull/16752 has been merged. 

Part of the PR are some additional docs, and I don't think this alone doesn't make the case for adding something to the release notes.

Issue #1738, opened 2023-01-05
Deprecate SHA1 based algorithms for SAML signatures


mposolda:
### Description

See https://github.com/keycloak/keycloak/pull/16273 comment for the details.

PR #1754, opened 2023-01-26
Algorithms are not links to click but properties to be set.

ahus1:
As external link validation fails due to a redirect to HTTPS, keep them as text.

> Error:    ExternalLinksTest.checkExternalLinks:26 Broken links (2):
		- http://www.w3.org/2000/09/xmldsig#rsa-sha1 (invalid redirect to https://www.w3.org/2000/09/xmldsig)
		- http://www.w3.org/2000/09/xmldsig#dsa-sha1 (invalid redirect to https://www.w3.org/2000/09/xmldsig)

Follow-up to #1738


ahus1:
@mposolda - this updates two properties to no longer be links. I see you added those recently, so I ask you for a review. 

From what I read from the docs, those links are not supposed to be clicked, but properties. 

There is still one more link failing, but that one is different, as that one should still be a link, but with https. I'll create another PR for that.

Please confirm. Thanks!

mposolda (approved):
@ahus1 Thanks for fixing this!

PR #1765, opened 2023-02-04
Replace misleading image

dreamcrash:
The original image has a User Policy but in reality it should have had a Client Policy. Related with the issue https://github.com/keycloak/keycloak/issues/19157


ahus1:
Thank you for reporting this and sorry for the late reply. This has now been fixed as part of issue keycloak/keycloak#16965 and PR keycloak/keycloak-documentation#1774

PR #1770, opened 2023-02-13
SPNEGO Warning

### LINKED ISSUE #16981: Docs are misleading about mixed domain Kerberos SPNEGO
### Before reporting an issue

- [x] I have reproduced the issue with the latest release
- [X] I have searched existing issues

### Area

docs

### Describe the bug

The docs [state](https://www.keycloak.org/docs/20.0.3/server_admin/#_kerberos):

```
5. If the browser has a Kerberos ticket from desktop login, the browser transfers the desktop sign-on information to Keycloak in header Authorization: Negotiate 'spnego-token'. Otherwise, it displays the standard login screen, and the user enters the login credentials.
```

That's not always true and is browser dependent.  Only if a web browser has been configured to participate in `Authorization: Negotiate` then the token will be sent.  If configured, but the token is invalid, then the standard login screen is displayed.

The missing part however, is if not configured, some browsers fall back to an NTLM dialog prompt.  This is often not what you want and difficult to work around.  At the moment Firefox latest (109.0.1) does behave as originally described and will not use NTLM by default (though I think it can be enabled client-side).  Chrome latest (110.0.5481.78) on Windows (but not Linux or MacOS) will prompt for NTLM by default.  It can be [disabled](https://community.microfocus.com/cyberres/netiq-access-management/accessmanager/f/nam_discussions/209874/kerberos-auth-can-the-ntlm-dialog-be-prevented), but requires control of every user's browser.

SPENGO is only useful in a tightly controlled Intranet, however it may be desirable to allow a single Keycloak cluster to serve a mixed environment where internal (Intranet) users are treated to SPENGO login and external users must use alternative means (username and password).

A good explanation of this issue can be found [here](https://knowledge.broadcom.com/external/article/202739/kerberos-fallback-to-form-chain-authenti.html).  The TLDR: Instead of adding a `Kerberos` www-authenticate [scheme](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml) which would allow servers to specify exactly what auth types are allowed, Microsoft created [Negotiate](https://www.ietf.org/rfc/rfc4559.txt) instead, which bundles Kerberos and NTLM together such that NTLM is an implied fallback (though browsers can disable NTLM to avoid it).

### Version

20.0.3

### Expected behavior

Docs should give clearer picture of SPNEGO behavior and support in Keycloak.  It's clumsy at best outside of an Intranet and if attempting a mixed domain environment beware no standard solution exists.   Detecting client IPs at the load balancer and redirecting to Keycloak using an idp_hint or [prompt=login](https://github.com/keycloak/keycloak/issues/8989) parameter may be one hack.  Another is providing an explicit separate link to try SPNEGO for users to select on the login page or in each app.

### Actual behavior

Docs leave out some important details and considerations.

### How to Reproduce?

Read docs.

### Anything else?

_No response_

slominskir:
Closes https://github.com/keycloak/keycloak/issues/16981

### DIFF
--- a/server_admin/topics/authentication/kerberos.adoc
+++ b/server_admin/topics/authentication/kerberos.adoc
@@ -16,6 +16,11 @@ A typical use case for web authentication is the following:
 . If using LDAPFederationProvider with Kerberos authentication support, {project_name} provisions user data from LDAP. If using KerberosFederationProvider, {project_name} lets the user update the profile and pre-fill  login data.
 . {project_name} returns to the application. {project_name} and the application communicate through OpenID Connect or SAML messages. {project_name} acts as a broker to Kerberos/SPNEGO login. Therefore {project_name} authenticating through Kerberos is hidden from the application.
 
+[WARNING]
+====
+The https://www.ietf.org/rfc/rfc4559.txt[Negotiate] www-authenticate scheme allows NTLM as a fallback to Kerberos and on some web browsers in Windows NTLM is supported by default.  If a www-authenticate challenge comes from a server outside a browsers permitted list then users may encounter an NTLM dialog prompt.  A user would need to click the cancel button on the dialog to continue as Keycloak does not support this mechanism.  This can happen if Intranet web browsers are not strictly configured or if Keycloak serves users in both the Intranet and Internet.  A https://github.com/keycloak/keycloak/issues/8989[custom authenticator] can be used to restrict Negotiate challenges to a whitelist of hosts.
+====
+
 Perform the following steps to set up Kerberos authentication:
 
 . The setup and configuration of the Kerberos server (KDC).
--- a/server_admin/topics/authentication/kerberos.adoc
+++ b/server_admin/topics/authentication/kerberos.adoc
@@ -18,7 +18,7 @@ A typical use case for web authentication is the following:
 
 [WARNING]
 ====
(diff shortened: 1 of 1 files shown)


slominskir:
**Note**: If Keycloak adds a conditional SPNEGO authenticator in the future the doc provided here would be updated to reference the built-in authenticator instead of the issue ticket discussing the custom one used to work around the NTLM issue.

See: https://github.com/keycloak/keycloak/pull/17064

mposolda (dismissed):
@slominskir Thanks! This is useful note to add

@andymunro Can you please review as well?

server_admin/topics/authentication/kerberos.adoc
```diff
@@ -16,6 +16,11 @@ A typical use case for web authentication is the following:
...
 . {project_name} returns to the application. {project_name} and the application communicate through OpenID Connect or SAML messages. {project_name} acts as a broker to Kerberos/SPNEGO login. Therefore {project_name} authenticating through Kerberos is hidden from the application.
 
+[WARNING]
+====
+The https://www.ietf.org/rfc/rfc4559.txt[Negotiate] www-authenticate scheme allows NTLM as a fallback to Kerberos and on some web browsers in Windows NTLM is supported by default.  If a www-authenticate challenge comes from a server outside a browsers permitted list then users may encounter an NTLM dialog prompt.  A user would need to click the cancel button on the dialog to continue as Keycloak does not support this mechanism.  This can happen if Intranet web browsers are not strictly configured or if Keycloak serves users in both the Intranet and Internet.  A https://github.com/keycloak/keycloak/issues/8989[custom authenticator] can be used to restrict Negotiate challenges to a whitelist of hosts.
```
andymunro:
```suggestion
The https://www.ietf.org/rfc/rfc4559.txt[Negotiate] www-authenticate scheme allows NTLM as a fallback to Kerberos and on some web browsers in Windows NTLM is supported by default.  If a www-authenticate challenge comes from a server outside a browsers permitted list, users may encounter an NTLM dialog prompt.  A user would need to click the cancel button on the dialog to continue as Keycloak does not support this mechanism.  This situation can happen if Intranet web browsers are not strictly configured or if Keycloak serves users in both the Intranet and Internet.  A https://github.com/keycloak/keycloak/issues/8989[custom authenticator] can be used to restrict Negotiate challenges to a whitelist of hosts.
```

andymunro (changes requested):
Thanks, just one change to suggest @slominskir

mposolda (approved):
@andymunro Thanks!

PR #1773, opened 2023-02-15
This should prob say "any" not "all"

slominskir:
See: https://github.com/keycloak/keycloak/pull/17064#discussion_r1107479351


slominskir:
See Also: https://github.com/keycloak/keycloak/discussions/10494

slominskir:
Also see different [section of docs](https://www.keycloak.org/docs/20.0.3/server_admin/#conditions-in-conditional-flows) that confirms it should be "any", not "all":

> If some executions included in the Conditional sub-flow evaluate as false, then the whole sub-flow is considered as Disabled.

mposolda (approved):
@slominskir Thanks

PR #1797, opened 2023-03-10
Update enforcer-overview.adoc to add 'by'

ericmiranda7:
missing 'by'

### DIFF
--- a/authorization_services/topics/enforcer-overview.adoc
+++ b/authorization_services/topics/enforcer-overview.adoc
@@ -25,7 +25,7 @@ If you are using any of the {project_name} OIDC adapters, you can easily enable 
 }
 ```
 
-When you enable the policy enforcer all requests sent by your application are intercepted and access to protected resources will be granted
+When you enable the policy enforcer all requests sent to your application are intercepted and access to protected resources will be granted
 depending on the permissions granted by {project_name} to the identity making the request.
 
 Policy enforcement is strongly linked to your application's paths and the <<_resource_overview, resources>> you created for a resource server using the {project_name} Administration Console. By default,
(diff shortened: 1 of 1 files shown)


ahus1:
@ericmiranda7 - thanks for creating this PR. I agree that the sentence is missing word. 

Looking at the context with the diagram above, the "application" is the the resource server ("It acts as a filter or interceptor in your application in order ..."). 

By adding "by" here, the sentence would be correct if the client would be the application, as it sends the requests. But as the application is the resource server, I suppose "to" is the word that is missing here.

If you agree, please update this PR. Thanks!

stianst (dismissed):

authorization_services/topics/enforcer-overview.adoc
```diff
@@ -25,7 +25,7 @@ If you are using any of the {project_name} OIDC adapters, you can easily enable
 }
 ```
 
-When you enable the policy enforcer all requests sent your application are intercepted and access to protected resources will be granted
+When you enable the policy enforcer all requests sent by your application are intercepted and access to protected resources will be granted
```
stianst:
```suggestion
When you enable the policy enforcer all requests sent to your application are intercepted and access to protected resources will be granted
```

stianst:
> @ericmiranda7 - thanks for creating this PR. I agree that the sentence is missing word.
> 
> Looking at the context with the diagram above, the "application" is the the resource server ("It acts as a filter or interceptor in your application in order ...").
> 
> By adding "by" here, the sentence would be correct if the client would be the application, as it sends the requests. But as the application is the resource server, I suppose "to" is the word that is missing here.
> 
> If you agree, please update this PR. Thanks!

You are correct, so changed this to `to`

